Issue #339, opened 2014-08-19
Readonly fields get displayed incorrectly in History view


dgaus:
I am trying to show users a read-only version of objects' history. The problem is, if I set any fields as read-only, then those show the current version of the object instead of the Version's. 
This seems to be because the render_revision_form method is using the ModelForm 'initial' argument, and then the fieldset.html template does {{ field.contents }}, which points to the original object.

I was thinking we could replace the line
```python
form = ModelForm(instance=obj, initial=self.get_revision_form_data(request, obj, version))
```
with a forloop that modifies obj to with the version data. Do you think this would work, or would it break anything else? Thanks

jmurty:
FWIW I have tackled a similar issue by:
* adding custom Javascript to the admin page to disable form fields on the client-side
* using the "strict" restore feature in PR #424 to enforce the read-only nature of history pages on the server-side, so user supplied form data is ignored altogether on revert

Not very elegant, but it does the job.

etianen:
Should be fixed in aa858f8ff53b50ea10be8b2ccca96ca993b95545

Try the latest master.

dgaus:
Large commit! looks nice. I wanted to test this, however after uninstalling and installing master (doing pip install https://github.com/etianen/django-reversion/archive/master.zip) I get 'ImportError: No module named reversion' when trying to run a django shell. Am I missing something obvious or is this a bug?

etianen:
Try pip install git+https://github.com/etianen/django-reversion.git

etianen:
And yes, it’s a fundamental change to how admin rollback is achieved. Before, it was a load of little hacks to smash the revision data into the admin form. Now, it uses a database savepoint to temporarily restore the DB to a point in time, render the page, then roll back the save point.

It’s a bit of an abuse of save points, but it seems to work in all DB engines. There are some caveats, such as it breaking MyISAM tables in MySQL (but that’s no longer the default table format in recent versions), and breaking for a transaction isolation of read uncommitted (which isn’t the default isolation level in any database anywhere). The benefit is that a huge number of fragile hacks are gone, and that a load of small things that didn’t work now work. :)

dgaus:
I think given the hackyness of the previous implementation this is a really good change. Cannot test it unfortunately! 

```
$ pip uninstall django-reversion
Uninstalling django-reversion:
  /home/vagrant/venv/lib/python2.7/site-packages/django_reversion-1.8.7-py2.7.egg-info
Proceed (y/n)? y
[...]

PR #340, opened 2014-08-19
Refs #339 Readonly fields get displayed incorrectly in History view

dgaus:
Instead of passing the Version parameters as initial, override them in the object itself


etianen:
Closing, as fixed in aa858f8ff53b50ea10be8b2ccca96ca993b95545

PR #352, opened 2014-09-11
fix for user lost when data been updated from outside the admin site

TommyU:
the user/comment of the reversion log will get lost when data been updated from outside the admin site.
and my test code which shows the bug is available at https://github.com/TommyU/test_reversion/


etianen:
Closing this. No recent activity, and the feature can be implemented in non-library code.

PR #356, opened 2014-09-24
Update django-versions.rst

ThorArakis:
Looks like 1.8.3 was the first version that required django 1.7+

### DIFF
--- a/docs/django-versions.rst
+++ b/docs/django-versions.rst
@@ -11,7 +11,7 @@ If you are using anything other than the latest release of Django, it is importa
 Django version  Reversion release
 ==============  =================
 1.7+            1.8.4
-1.6+            1.8.4
+1.6+            1.8.2
 1.5.1+          1.7.1
 1.5             1.7  
 1.4.4+          1.6.6


jedie:
Seems that v1.8.2 is the last one for Django v1.6.x isn't it?

See: https://github.com/etianen/django-reversion/compare/release-1.8.2...release-1.8.3

A compatibility table would be good ;) Or does this anywhere exists?

etianen:
The current release (1.8.5) should still support Django 1.6.x.

http://django-reversion.readthedocs.org/en/latest/django-versions.html

jedie:
Thanks for the link! Maybe a good place of a link to is is in the README under https://github.com/etianen/django-reversion#upgrading ?

> The current release (1.8.5) should still support Django 1.6.x.

No.
I think this commit it the problem: https://github.com/etianen/django-reversion/commit/1f265dbe1fb499032954b4f589816558d03fd9b0 and will result in ```ImportError: cannot import name migrations``` (As sayed above: https://github.com/etianen/django-reversion/pull/356#issuecomment-57344919 )

it's the switch from ```south``` to ```django.db.migrations``` which is new in django 1.7.x

etianen:
It totally works in 1.6! I have proof! :P

https://travis-ci.org/etianen/django-reversion

Try upgrading to south 1.0.

PR #358, opened 2014-10-05
South migrations did not support custom user models

thedrow:
For some reason only the initial south migration supported a custom user model.
I added it to all other migrations and added support for user models without integer primary keys.

[Review on Reviewable](https://reviewable.io/reviews/etianen/django-reversion/358)


etianen:
South migrations are removed from the project now, so this is no longer relevant.

PR #375, opened 2014-12-16
Support for MySQL utf8mb4

alexhayes:
As outlined in http://dev.mysql.com/doc/refman/5.5/en/charset-unicode-upgrading.html

> InnoDB has a maximum index length of 767 bytes, so for utf8 or utf8mb4 columns, you can index a maximum of 255 or 191 characters, respectively. If you currently have utf8 columns with indexes longer than 191 characters, you will need to index a smaller number of characters. In an InnoDB table, these column and index definitions are legal:

This change ensures those using the InnoDB storage engine in MySQL with utf8mb4_* collations can use reversion.

### DIFF
--- a/src/reversion/migrations/0002_auto_20141216_1509.py
+++ b/src/reversion/migrations/0002_auto_20141216_1509.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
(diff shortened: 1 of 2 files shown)


alexhayes:
Yes, but only if you're using the DYNAMIC and COMPRESSED row formats, but I am not and I imagine that most won't be either because the default is COMPACT (since MySQL 5.0.3).

etianen:
`Seems legit. I'm pulling now.

PR #381, opened 2015-01-28
fixed #380: Versions passed through by reversion.post_revision_commit ha...

### LINKED ISSUE #380: Versions passed through by reversion.post_revision_commit have no primary key
Odd behavior:

```py
# Setup this signal, try saving a tracked object, and check your test server output
@receiver(reversion.post_revision_commit)
def on_revision_commit(revision, versions, instances, **kwargs):
    print(versions[0].pk)
    # pk = None

    # then if I save the object to give it a pk, it saves a duplicate in the db
    versions[0].save()  # go check the db, you'll see the duplicate
    print(versions[0].pk)

```

joelarson:


etianen:
Thanks!

PR #382, opened 2015-01-31
Update admin.py to use inline.get_queryset

adonm:
Use inline.get_queryset instead of inline.queryset to correctly revert modeladmins with tabular/stacked inlines.


etianen:
Good spot. Ta!

PR #386, opened 2015-02-10
Fix deprecated import of GenericForeignKey

khakulov:
Importing of django.contrib.contenttypes.generic is deprecated since Django 1.7. https://docs.djangoproject.com/en/1.7/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericForeignKey

    Deprecated since version 1.7:
    This class used to be defined in django.contrib.contenttypes.generic. Support for importing from this old location will be removed in Django 1.9.

### DIFF
--- a/src/reversion/models.py
+++ b/src/reversion/models.py
@@ -3,7 +3,7 @@
 from __future__ import unicode_literals
 
 from django.contrib.contenttypes.models import ContentType
-from django.contrib.contenttypes import generic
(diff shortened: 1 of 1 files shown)


etianen:
This will break Django 1.6 support. For the time being, until 1.8 is released, I'd like to keep supporting 1.6.

I'll keep this one on hold until then. But still, thanks for contributing, and it will be merged soon!

PR #388, opened 2015-03-12
Fix Django 1.8+ warnings by importing from django.contrib.admin.utils (plural)

### LINKED ISSUE #384: RemovedInDjango19Warning: django.contrib.admin.util
I tried to run my project with Django 1.8. I got the following warning:

    RemovedInDjango19Warning: The django.contrib.admin.util module has been renamed.
    Use django.contrib.admin.utils instead.
    "Use django.contrib.admin.utils instead.", RemovedInDjango19Warning)

admin.py L:12
```python
from django.contrib.admin.util import unquote, quote
```

The other question for myself was, why the  ```admin.py ``` was imported in my project, where i never use django admin interface/modules. Why we need to import  ```VersionAdmin ``` class in ```__init__.py```???

blag:
Closes #384.

### DIFF
--- a/src/reversion/admin.py
+++ b/src/reversion/admin.py
@@ -9,7 +9,10 @@ from django.db import models, transaction, connection
 from django.conf.urls import patterns, url
 from django.contrib import admin
 from django.contrib.admin import helpers, options
-from django.contrib.admin.utils import unquote, quote
+try:
+    from django.contrib.admin.utils import unquote, quote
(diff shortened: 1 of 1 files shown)


etianen:
This breaks backwards compatibility with 1.6. Until 1.8 has been released, I'd like to keep this pull request on hold.

Alternatively, a patch that performs a conditional import would be merged right now.

blag:
Good point. Fixed.

PR #394, opened 2015-04-24
Update admin.py to avoid RemovedInDjango19Warning

niknokseyer:
RemovedInDjango19Warning: django.contrib.contenttypes.generic is deprecated and will be removed in Django 1.9. Its contents have been moved to the fields, forms, and admin submodules of django.contrib.contenttypes.


etianen:
Thanks!

PR #395, opened 2015-04-29
Update admin.py to avoid RemovedInDjango19Warning

BATCOH:
#394 does not work

### DIFF
--- a/src/reversion/admin.py
+++ b/src/reversion/admin.py
@@ -13,10 +13,11 @@ try:
     from django.contrib.admin.utils import unquote, quote
 except ImportError:  # Django < 1.7
     from django.contrib.admin.util import unquote, quote
-try:    
-    from django.contrib.contenttypes.generic import GenericInlineModelAdmin, GenericRelation
+try:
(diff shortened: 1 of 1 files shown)


etianen:
Can you clarify "doesn't work"?

BATCOH:
django.contrib.contenttypes.generic will be removed in Django 1.9 and generate RemovedInDjango19Warning in 1.8.
In Django 1.8+ GenericInlineModelAdmin moved to django.contrib.contenttypes.admin and GenericRelation to django.contrib.contenttypes.fields

To avoid warnings and errors we need to try import them from new place first and fallback to django.contrib.contenttypes.generic on ImportError (for older versions of Django).

#394 have few bugs with that:
import from django.contrib.contenttypes.generic first (and we still get RemovedInDjango19Warning in 1.8)
And on ImportError we try to get both GenericInlineModelAdmin and GenericRelation from django.contrib.contenttypes.admin (but GenericRelation located in django.contrib.contenttypes.fields, so we will get another ImportError in 1.9)

P.S. Sorry for possible writing mistakes, English is not my primary language.

PR #401, opened 2015-05-06
Fix for deleterevisions on Py3

davidfsmith:
`./manage.py deleterevisions` should now work on both Python2 and 3 this change resolves the error below when running on Python3

```
choice = raw_input("Are you sure you want to delete theses revisions? [y|N] ")
NameError: name 'raw_input' is not defined
```


etianen:
Looks good. Thanks!

I'm really loving how the Django community appears to be finally switching over to Python 3. :)

davidfsmith:
Thanks.

It’s about time to be honest ;-)

PR #406, opened 2015-05-21
Replaced/removed some old stuff

claudep:


src/reversion/management/commands/createinitialrevisions.py
```diff
@@ -1,14 +1,22 @@
...
 from optparse import make_option
+try:
+    from collections import OrderedDict
+except ImportError:  # For Python 2.6
+    from django.utils.datastructures import SortedDict as OrderedDict
+
+try:
+    from django.apps import apps
+except ImportError:  # For Django < 1.7
+    from django.db import models as apps
```
etianen: This feels like a fallback that isn't correct, as the two modules aren't equivalent. We should really be specifically importing get_models() and get_apps() using from django.(apps/models) import get_models, get_apps.

src/reversion/revisions.py
```diff
@@ -8,12 +8,17 @@
 from weakref import WeakValueDictionary
 import copy
 
+try:
+    from django.apps import apps
+except ImportError:  # For Django < 1.7
+    from django.db import models as apps
+
```
etianen: This feels like a fallback that isn't correct, as the two modules aren't equivalent. We should really be specifically importing get_models() and get_apps() using from django.(apps/models) import get_models, get_apps.

etianen:
Aside from my notes about get_apps() and get_models(), this all looks good. If you could make the change, I'll pull. :)

src/reversion/management/commands/createinitialrevisions.py
claudep: I didn't look in details about differences, but it seems at least in Django 1.8 they were equivalent enough to justify some aliasing:
https://github.com/django/django/blob/stable/1.8.x/django/db/models/__init__.py#L62
https://github.com/django/django/blob/stable/1.8.x/django/db/models/loading.py

etianen: Yes, but django.db.models includes other things like models.Q, models.Max, which aren’t present in django.apps.

claudep: Ah, I see your point now. I'll directly import the methods instead.

claudep:
FYI the PR has been updated.

etianen:
Lovely, it's merged! Thanks for the fixes. :)

claudep:
Thanks for merging. One of the latest deprecation warning on Django 1.8 is the usage of `get_formsets` in admin.py. However, this is a bit harder to fix as some knowledge of that code is required to update it wisely...

etianen:
What specific warning are you getting for get_formsets?

claudep:
```
.../local/lib/python2.7/site-packages/reversion/admin.py:329: RemovedInDjango19Warning:
ModelAdmin.get_formsets() is deprecated and will be removed in Django 1.9. Use ModelAdmin.get_formsets_with_inlines() instead.
  for FormSet, inline in zip(self.get_formsets(request, obj), self.get_inline_instances(request)):
```
See https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_formsets

Issue #391, opened 2015-04-08
Create then delete within view when using middleware


d-mcc-af:
Hey, thanks for your work. 

When using the reversion middleware, if I create and save an object within a view, then immediately delete it, I get the following stack.

```
Stacktrace (most recent call last):

  File "django/core/handlers/base.py", line 199, in get_response
    response = middleware_method(request, response)
  File "reversion/middleware.py", line 37, in process_response
    self._close_revision(request)
  File "reversion/middleware.py", line 28, in _close_revision
    revision_context_manager.end()
  File "reversion/revisions.py", line 176, in end
    in manager_context.items()
  File "reversion/revisions.py", line 175, in <genexpr>
    for obj, data
  File "reversion/revisions.py", line 616, in <lambda>
    version_data = lambda: adapter.get_version_data(instance, self._revision_context_manager._db)
  File "reversion/revisions.py", line 100, in get_version_data
    object_id_int = int(obj.pk)
```

etianen:
Can you include the full stack trace? It looks like the bottom is missing.

chrismcmacken:
I'm having the same issue, here's the full stack trace. It looks like @d-mcc-af is using the middleware, but I'm manually flagging the revisions in each method. The error doesn't occur every time, sometimes it works fine and other times it explodes with this error. 

Traceback:
File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/site-packages/django/views/decorators/csrf.py" in wrapped_view
  57.         return view_func(*args, **kwargs)
File "/usr/local/lib/python2.7/site-packages/rest_framework/viewsets.py" in view
  85.             return self.dispatch(request, *args, **kwargs)
File "/usr/local/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  407.             response = self.handle_exception(exc)
File "/usr/local/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  404.             response = handler(request, *args, **kwargs)
File "/usr/local/lib/python2.7/site-packages/django/db/transaction.py" in inner
  394.                 return func(*args, **kwargs)
File "/usr/local/lib/python2.7/site-packages/reversion/revisions.py" in do_revision_context
  305.                 return func(*args, **kwargs)
File "/usr/local/lib/python2.7/site-packages/reversion/revisions.py" in __exit__
[...]

PR #408, opened 2015-05-27
Properly handles a scenario when an instance's PK is None

kpantic:
When copying instances in Django, the standard way of doing it is setting an object's PK as None. As of some versions of Django, when a related field is assigned, a save event is triggered and if that assignment was done on the object that was being copied and had a PK of None it created a conflict with django-reversion.

When reversion tried to save changes done to the object, it could not locate the PK of the object, raising an Exception since the int() conversion of None returns a ValueError.

This PR fixes it by checking that the object that will be saved in the revision has a PK that is not None and thus skips it.

### DIFF
--- a/src/reversion/revisions.py
+++ b/src/reversion/revisions.py
@@ -174,6 +174,7 @@ class RevisionContextManager(local):
                                 (obj, callable(data) and data() or data)
                                 for obj, data
                                 in manager_context.items()
+                                if obj.pk
                             ),
                             user = self._user,
                             comment = self._comment,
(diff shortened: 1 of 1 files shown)


src/reversion/revisions.py
```diff
@@ -174,6 +174,7 @@ def end(self):
                                 (obj, callable(data) and data() or data)
                                 for obj, data
                                 in manager_context.items()
+                                if obj.pk
```
etianen: I think this should be:

``` py
if obj.pk is not None
```

etianen:
Nice. I think this will also fix #391.

I've made a comment on your implementation that I think will improve it. If you could make the change, I'll pull.

kpantic:
Yeah I noticed it as soon as I made the PR. I'll make the change.

kpantic:
I made the change, so it's ready to be merged.

etianen:
Great, thanks!

PR #410, opened 2015-06-01
Test that creating and deleting instance does not create version

podloucky-init:
This verifies that #391 has indeed been fixed.

Adds a test that verifies that creating and deleting a model instance
within a reversion does not create a new version.


etianen:
Lovely!

PR #418, opened 2015-06-25
Fix deleterevisions' user choice prompt to work with Python 2

jmurty:
When prompting the user for a y/n choice, the deleterevisions command
was failing for me in Python 2.7.9 because it was falling back to use
input() instead of raw_input().

This fallback is appropriate for Python 3 in which raw_input() was
replaced by input(), but should never be done in Python 2 because there
the input() function evaluates the user input as code.

I am not really sure why the attempted Python 2/3 recognition technique
of assigning the raw_input function to itself fails to work correctly
for me, but this restructured version should accomplish the same thing
with less chance of incorrectly assuming it is in Python 3.

### DIFF
--- a/src/reversion/management/commands/deleterevisions.py
+++ b/src/reversion/management/commands/deleterevisions.py
@@ -1,12 +1,13 @@
 from __future__ import unicode_literals
 
-import datetime, operator, sys
+import datetime, operator
 from optparse import make_option
 
 from django.core.management.base import BaseCommand
 from django.core.management.base import CommandError
 from django.db.models import Q, Count
 from django.contrib.contenttypes.models import ContentType
+from django.utils.six.moves import input
 
 from reversion.models import Revision, Version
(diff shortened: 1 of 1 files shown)


etianen:
A better way to do this might be to use six in the head of the module.

``` py
from django.utils.six.moves import input
```

jmurty:
Hi @etianen I agree using six.moves would be better, but the six compatibility layer it isn't used elsewhere in the codebase and I didn't want to introduce a new dependency. It should be fine though since `django.utils.six` has been available since [Django 1.4.2](https://docs.djangoproject.com/en/1.8/topics/python3/#writing-compatible-code-with-six) (earlier than I thought)

etianen:
Full marks for trying to fit with the existing code style! I like this new approach better. Merged!

jmurty:
Thanks!

PR #423, opened 2015-07-07
Add revision manager that auto-registers models in follow relationships

jmurty:
Add a new `AutoRegisterRevisionManager` implementation of `RevisionManager` that will automatically register unregistered model classes encountered via `follow` relationships when saving a new revision.

The auto-registration feature is intended to help add versioning support to a CMS where there can be a complex object relationship model and a set of content types models that grows over time, where manually adding explicit `register` calls for every model is repetitive and easy to forget.

PR also includes a `RevisionManager` change so on init you can set a default adapter class to use when registering models, making it easier to use a custom adapter for all model registrations without needing to provide the adapter as an argument with every `register` call.

For context, this work is being used as part of an effort to add reversion support to django-fluent-pages, see use of `AutoRegisterRevisionManager` at bottom: https://github.com/ixc/django-fluent-pages/blob/reversion_support/fluent_pages/integration/django_reversion/utils.py

### DIFF
--- a/src/reversion/revisions.py
+++ b/src/reversion/revisions.py
@@ -7,6 +7,7 @@ from functools import wraps, reduce, partial
 from threading import local
 from weakref import WeakValueDictionary
 import copy
+import logging
 
 try:
     from django.apps.apps import get_model
@@ -24,6 +25,9 @@ from django.db.models.signals import post_save
 from django.utils.encoding import force_text
 
 from reversion.models import Revision, Version, has_int_pk, pre_revision_commit, post_revision_commit
+
+
+logger = logging.getLogger(__name__)
 
 
 class VersionAdapter(object):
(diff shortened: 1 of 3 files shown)


etianen:
This feels like quite a specialised use-case.

In the django admin integration that django-reversion ships with, it performs autodetection of followed relationships when it initializes. Couldn't the admin registration for django-fluent-pages support something similar?

jmurty:
I will need to look further into how the Django admin integration auto detection of followed relationships works.

This approach might work for our use-case if we can find and register all the polymorphic models that could be associated with the followed relationships of a page.

jmurty:
Deprecating this PR. I am able to auto-register all the models we need as part of the admin auto-registration process which isn't pretty, but is probably cleaner than this approach.

PR #424, opened 2015-07-07
Add support for "strict" revert to Django admin integration

jmurty:
Add support for "strict" revert/restore in Django admin integration if the new `VersionAdmin.strict_revert` class attribute is True (defaults to False)

When strict revert is enabled and you revert/restore version data via the built-in Django admin integration, the *exact* prior version will be restored and any modifications made in the submitted admin form data will be ignored.

This change makes it possible to safely revert complex object graphs without needing to guarantee that *every* piece of data in the prior version is also properly represented in a single admin form, which isn't always possible for situations like a CMS where one object may have multiple translatable fields that are managed by loading different translations into a single base admin form depending on the current language of interest.

For context, this work is being used as part of an effort to add reversion support to django-fluent-pages: https://github.com/ixc/django-fluent-pages/tree/reversion_support

### DIFF
--- a/src/reversion/admin.py
+++ b/src/reversion/admin.py
@@ -276,7 +276,18 @@ class VersionAdmin(admin.ModelAdmin):
             # If configured to restore exact version data, do so without
             # processing any submitted form data which may have been altered...
             if getattr(settings, 'REVERSION_ADMIN_STRICT_REVERT', False):
-                version.revision.revert(delete=True)
+                if getattr(settings,
+                           'REVERSION_ADMIN_STRICT_REVERT_WITH_PRE_DELETE',
+                           False):
+                    # Brutal hack to remove obsolete relationships from the
+                    # current in-database object by deleting the entire object.
+                    # This is necessary in cases where the versioned object
+                    # graph is too complex for reversion's inbuilt
+                    # delete-on-revert feature to handle.
+                    obj.delete()
+                    version.revision.revert()
+                else:
+                    version.revision.revert(delete=True)
                 new_object = model.objects.get(pk=object_id)
                 was_reverted = True
 
--- a/src/reversion/admin.py
+++ b/src/reversion/admin.py
@@ -6,6 +6,7 @@ from functools import partial
 
 from django import template
 from django.db import models, transaction, connection
+from django.conf import settings
 from django.conf.urls import patterns, url
 from django.contrib import admin
 from django.contrib.admin import helpers, options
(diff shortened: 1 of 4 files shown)


etianen:
I can see the use for this, but I'm not sure that this is entirely the correct approach.

1. Rather than introduce a new django setting, I think it might be better to have it as an attribute of the VersionAdmin class.

2. It's a bit unintuitive to display a form to the user, but have it ignore any changes made! This feels like it needs a read-only preview of the version to be restored.  How that would be created is an open question.

jmurty:
Thanks for the feedback.

1. Moving the Django setting into a VersionAdmin class attribute makes sense, I will make that change.

2. We are handling the read-only nature of the revert/restore pages by applying Javascript to disable the form fields client-side. I looked into doing this on the server-side but it looked quite involved, and for our CMS would leave a lot of dynamic form editing UI "junk" we would still need to strip out anyway. Would you be interested in having some basic Javascript support for disabling the form in this project?

etianen:
Javascript seems wrong. I can see how a complicated CMS UI would require it, but the basic Django admin shouldn't.

How about this as an approach? Override `get_readonly_fields()`, and within this method get the list of form fields from `get_fieldsets()`, flatten it using `flatten_fieldsets()`. This will set all form fields to read-only.

The trick is getting it to work for inline model formsets, of course. Then again, for strict revert, those could just be omitted? It's lazy, but strict revert implies that your data transcends the inline formsets anyway.

I'm wondering whether an approach like this shouldn't be the default. The current hacking of the inline formsets is horrible, and a completely different mechanism for reverting that the low-level API. A cleaner approach might be to show a read-only no-inline view of the data, allow the developer to easily override it for a nicer read-only view, and get rid of all the formset hacking for good.

etianen:
Of course, a big change like this would mean a new major release. Now that I think about it, I'm increasingly in favour of this approach. It's much better for long-term maintainability and fixes some long-standing issues with the admin integration.

jmurty:
From my perspective a read-only view of the data makes a lot of sense, as opposed to allowing the user to edit and restore in the one operation, so making this major change in default behaviour in a new version of reversion sounds reasonable.

However, please don't scrap the admin inline formsets data population altogether! The inlines may be "hacky" – they actually work pretty well for us :) – but in our case an admin page without inlines would be next to useless since all the content of our CMS pages is "plugin" based and only appears in the inlines.

The strict restore feature makes it much less important for the admin page content to be entirely accurate and complete – it is enough for the admin page to show a reasonable approximation of the data that will be restored. Maybe switching to a strict restore as the default and making it clear that the data shown on admin pages (including inlines) is a best-effort only would sidestep some of the maintenance issues with the admin integration? Or at least, with users' expectations for it.

etianen:
To be fair, a read-only set of inlines would be a lot less hacky. Maybe that’s the compromise.

jmurty:
I should add that while I agree that using Javascript to disable the fields client-side feels pretty wrong, the Javascript approach is quite effective, relatively easy, and could work well for standard Django admin presentations like it does for our use-case with a complex dynamic admin UI.

My suggestion would be to make strict restore available as an optional feature now with the Javascript client-side hack to produce reasonable behaviour by default – quick and dirty as it is – then use that as an example for the proposed new default behaviour once the server-side can be sorted out.

jmurty:
I could probably have a go in the next couple of weeks at making the base and inline fields disabled/read-only on the server-side. I actually made a start down this path originally, before I realised how difficult it was likely to be (and that the Javascript hack was good enough for us)

etianen:
Unfortunately, making it optional with a javascript hack just means more code to maintain in the short-term. I don’t think the server-side solution would be much work.

I’ve have a think about it over the weekend. If this still seems like a good idea, I’d have a go at getting this working in a separate branch next week.

etianen:
(I’ve done dynamic read-only stuff before, and the code is still hanging around in some project somewhere. It’s not too bad.)

jmurty:
That sounds great thanks @etianen, I appreciate your efforts on this potentially disruptive PR, whatever you decide.

etianen:
Disruptive, for sure, but it’s started something that will make the codebase much better in the long-run. :)

etianen:
Have a look at this:

https://github.com/etianen/django-reversion/tree/admin-transactions

It's a variant of the idea we discussed. It uses a hybrid approach of using the normal admin widgets to display the revision data, but also performs a revert(delete=True) to revert any objects not in the admin forms.

In order to achieve this, I've (ab)used database transactions to actually revert the revision prior to rendering the form, then rolling back the transaction post-rendering. This has the advantage in that the data you're rendering the form with is completely consistent, without requiring any of the previous hacks to smoosh the revision data in.

Take a look, and let me know what you think. It's still a bit WIP, so anything you spot would be useful to know.

jmurty:
This seems like a really promising approach. It feels daring to use DB transactions like this, but if it works correctly then it makes for a much cleaner approach overall. 

I am now considering taking the same approach to allow viewing of versioned pages in full fidelity, rather than having to fake it (which was going to be complex for our CMS page objects). Maybe via some middleware...

I have only read through the code so far, not run it, but I have one request: Can you move the `version.revision.revert(delete=True)` call into a method so it can be overridden? We need to do so when reverting our unwieldy page objects that the inbuilt delete doesn't fully handle

jmurty:
I can confirm that the *admin-transactions* branch worked well for our project as a drop-in replacement for this strict versioning PR.

The only minor issue I have found is that when you restore or revert an item, the history comment lists all the field changes as for a standard version save, instead of stating "Reverted to previous version, saved on ..."

etianen:
I’m not sure that transactions were meant to be abused like this, but it works well. It’s a case of trading a lot of pervasive, smaller and incomplete hacks for a giant thermonuclear “reset the world” hack. Thankfully, database transactions and save points seem up to the task.

What aspect of version.revision.revert(delete=True) isn’t sufficient? If it’s something general, it could be fixed reversion-side.

I’ll fix the change message for the admin too.

etianen:
The change message update is in, and I’ve had a manual test everything, as well as the automated tests. I’m pretty happy with this.

I was thinking about making a context manager to allow a revision to be rolled back within a limited block of code, which sounds like you’d be able to use for your own preview functionality.

jmurty:
Thanks @etianen for finding a much better approach to enable "strict" restores. Closing this PR

PR #425, opened 2015-07-07
Improve support for versioning polymorphic models in Django admin integration

jmurty:
Improved handling of polymorphic models in Django admin:

* include `ct_id` context type URL parameter in recover listing and detail form when appropriate
* recognise related version data relevant to inline forms in edge-case situations where the version's apparent content type does not match the actual class that was versioned.

For context, this work is being used as part of an effort to add reversion support to django-fluent-pages: https://github.com/ixc/django-fluent-pages/tree/reversion_support

### DIFF
--- a/src/reversion/admin.py
+++ b/src/reversion/admin.py
@@ -370,7 +370,7 @@ class VersionAdmin(admin.ModelAdmin):
                         "has_delete_permission": self.has_delete_permission(request, obj),
                         "has_file_field": True,
                         "has_absolute_url": False,
-                        "form_url": mark_safe(request.path),
+                        "form_url": mark_safe(request.get_full_path()),
                         "opts": opts,
                         "content_type_id": ContentType.objects.get_for_model(self.model).id,
                         "save_as": False,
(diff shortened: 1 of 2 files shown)


etianen:
Could you explain the situation under which a "version's apparent content type does not match the actual class that was versioned"?

That sounds like a bug that should be fixed higher up. Shouldn't a version's content type always match the actual class that was versioned?

jmurty:
The edge-case prompting our work-around for inexact `ContentType`s is for some dynamically-generated classes (see bottom): https://github.com/edoburu/django-fluent-contents/blob/master/fluent_contents/plugins/markup/models.py

Basically, we have dynamically-generated class variants of the base `MarkupItem` such as `MarkdownMarkupItem` and `TextileMarkupItem`, but when you look up their ContentType you get the `MarkupItem` class which does not match the more exact/correct version (e.g. `MarkdownMarkupItem`) set on the `Formset.model` the method matches against.

Admittedly this is all pretty unusual stuff, and might only ever be useful for this particular CMS. Maybe the best option is to move the class comparison into a method so we can easily override it to handle our use-case? Or we could just override the `get_related_versions` implementation wholesale and leave the original unchanged, which is a bit messier but only for us ;)

jmurty:
This PR is deprecated, and has been replaced by the much simpler #438

PR #427, opened 2015-07-15
Fix typos: then -> than

tdruez:


etianen:
Yuck! Good spot!

tdruez:
Thanks!

Issue #431, opened 2015-07-17
Consider moving diff_match_patch to install_requires


pydanny:
During installation it didn't get the `diff_match_patch` dependency. I addressed that by adding it to my requirements directory, but it might be easier for coders to use the diff features if this was automatically installed.

pydanny:
For what it's worth, I'm on `pip 7.1.0`.

LegoStormtroopr:
Top my knowledge diff-patch is an optional extra, not a requirement.

> django-reversion comes with a number of helper functions that make generating diffs easy. They all rely on the google-diff-match-patch library, so make sure you have this installed before trying to use the functions.

From: https://django-reversion.readthedocs.org/en/latest/diffs.html

etianen:
Correct. I'm afraid that the diff match patch library helpers are not used by a particularly large chunk of the the user base. They're definitely an optional requirement. :)

PR #432, opened 2015-07-17
Moved diff_match_patch to install_requires

pydanny:
If #431 is accepted, this is the patch.


etianen:
As per the discussion in #431, I'm closing this.

PR #434, opened 2015-07-24
Czech translation update and fixes

cuchac:


etianen:
I'll take your word that these are improvements! :P

Thanks!

PR #436, opened 2015-07-29
Fix instructions for running tests in README.rst

jmurty:
Recent changes to the test-running procedure were not reflected in the README.

I assumed the process used currently for TravisCI is the correct one, but there may be a better way?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,9 +58,8 @@ commands:
 
 ::
 
-    $ pip install django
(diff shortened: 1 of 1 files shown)


jmurty:
NOTE: this should pass TravisCI tests once PR #437 is integrated

etianen:
Thanks! I'm an idiot! :P

PR #437, opened 2015-07-29
Fix argument ordering for get_unique_for_object deprecation warning

jmurty:

PR #438, opened 2015-07-29
Add content type ID to links on recover list page for polymorphic items

jmurty:
If a deleted item listed on the recover list page is a polymorphic
type -- as identified with a 'polymorphic_ctype' value in the version
data dict -- add the value as a '?ct_id' parameter on the item's link
to inform the recover page detail view of the item's polymorphic type.

### DIFF
--- a/src/reversion/templates/reversion/recover_list.html
+++ b/src/reversion/templates/reversion/recover_list.html
@@ -26,10 +26,12 @@
                     </thead>
                     <tbody>
                         {% for deletion in deleted %}
+                            {% with ct_id=deletion.field_dict.polymorphic_ctype %}
                             <tr>
-                                <th scope="row"><a href="{% url opts|admin_urlname:'recover' deletion.pk %}">{{deletion.revision.date_created}}</a></th>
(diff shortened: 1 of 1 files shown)


jmurty:
NOTE: this should pass TravisCI tests once PR #437 is integrated

etianen:
Does this ct_id correspond to anything in Django core? Or is it a specific variable set by your CMS?

I'd be reluctant to put this into the template if it's just for your CMS, since it can easily be implemented with a project-specific template provided by your CMS.

jmurty:
The `ct_id` isn't part of Django core but is used/required by the *django-polymorphic* admin for polymorphic models where the type isn't discoverable from the URL path, see for example: https://github.com/chrisglass/django_polymorphic/blob/master/polymorphic/admin.py#L321

I'm not sure how widely polymorphic models are used with reversion. We could certainly use a project-specific template override, though we would need to duplicate most of the *recover_list.html* page.

etianen:
Tricky. I wonder if there's a way that we could make the template more easy
to override in this case, or some useful hook that could be added to the
admin?

jmurty:
The only way I can think of to handle this more cleanly would be to add a "url_params" `block` to the end of the link in reversion's template, which would make customisation easier but is still pretty awful.

For now I have just overridden the "content" block in a customised template, so we can again use the official release.

If anyone can think of a better way of handling this situation let us know.

etianen:
Closing this for lack of activity. Please feel free to re-open!

PR #440, opened 2015-08-03
Document deleterevisions command

cuchac:
I was going to implement functionality to delete old revisions and suddenly discovered its already there. This is my attempt to document the 'deleterevisions' command slightly.


etianen:
Thanks! Very negligent of me...

PR #443, opened 2015-08-07
Support pre_delete signal in admin

cuchac:
Without this change 'pre_delete' eager signal is not creating any revision.
I'm not using version middleware to track only admin changes. With this change and eager_signal=[pre_delete] is delete/undelete of model without initial version working.


etianen:
Creating revisions on delete isn't a core feature of django-reversion. If you want this functionality, you'll have to subclass the VersionAdmin and add it yourself.

PR #447, opened 2015-08-26
Change order of `available_versions` in example

siamalekpour:
`reversion.get_for_object()` should return versions in a reversed-chronological order. So, `available_versions[0]` would reflect the latest change.


etianen:
Good spot. Thanks for updating the docs.

PR #449, opened 2015-09-06
i18n: Translate missing German localization string

lutoma:
Pretty straight and forward. This German translation string was missing, the PR adds it.


etianen:
Lovely! :D

PR #452, opened 2015-09-17
Updated index to use "manage.py migrate"

IanLee1521:
This replaces "manage.py syncdb" which is deprecated and will be removed in Django 1.9


etianen:
Good point!

Issue #465, opened 2015-11-21
Problem with proxy models and inlines


RamezIssac:
Use case: model and a sub model, both are proxies, created and edited via admin using inline,
Problem: The history view show only changes done to the parent model.
in other terms:
Recorded changes for proxy models within inlines are not visible on the history view

During investigation, i noticed in the version table, changes recorded foreignkey content_type to concrete model.  (changes in parent proxy are recorded pointing to concrete model, same with the child model)
while in django_admin_log -as example- ,  changes (in parent or child) are recorded pointing to the parent proxy model, which makes it easier to retrieve.

How can i go about solving the Problem? any hints where to start ?

RamezIssac:
Well well, i couldn't reproduce it in reversion tests ... :-)

Could you elaborate on how the versions get collected for the history view,like a guide to reversion internals if you may 

Thanks.

etianen:
The versions are collected by following a reverse foreign key, which is set up in the `follow` argument of `register`.

So, if you call `register(YourModel, follow=("something_set",))`, then whatever is returned by following that model property is used to populate the inlines.

etianen:
I've just updated the tests to use a proxy parent and a proxy child model. Still can't reproduce, I'm afraid.

RamezIssac:
I confirm, issue not reproduced. That's interesting :)
Will give my project some debugging to know where the deviation came from and get back to you.

Grateful for your help and patience

etianen:
No worries! I'd be very interested to track this down.

The latest django-reversion uses database transactions to roll back
revisions for preview, which should make it work with extremely esoteric
examples of proxy models. I'd be very interested if there is a way to break
it!

RamezIssac:
GOT IT !!
The issue appear when the Base Models (not proxy) are not used on their own , so an error page appear suggesting to register (once for the inline first, then for the parent), so i do and register them explicitly,
and UNLESS the parent base model is registered using the follow=['children'], the issue will appear.

And that's also the only difference between your added test, and the PR #471 

Phewwww :-)

etianen:
So you just need to update your register code, and it works?

Phew! :D

etianen:
Anyway, thanks for taking the time to put together actual test cases to
help debug the issue. If only all issues were like that... :P

RamezIssac:
Well.... 
What do you think ? 
Shouldn't the automatic registration handle such case and follow what the actual inlines says (as long as base models are registered) ? 
[...]

PR #471, opened 2015-12-07
PR demonstrating #465 (proxy model and inline)

RamezIssac:
Issue #465 
Problem with proxy (models and inlines)

Currently failed test demonstrates that when looking at initial version (and in fact any other version), the history view shows the "current" inlines , not the inline version associated with this reversion history.

*don't know exactly how some old commits to update my fork appeared in this PR. kindly ignore them.*

Thanks.


etianen:
So the issue can be reproduced by:

1) Register an inline admin class for a proxy model.
2) Create a revision with no children
3) Create a revisions with children
4) Load revision 1
5) Admin site shows that inlines

I've made a minimal test case for it here: c86bb87e64e7f3952bd53f4888cab19e1e5882cd

This test case is currently passing. I'm afraid that your pull request confounds the issue by using explicit registration and by proxying a model that's not otherwise used in admin tests.

RamezIssac:
Thank you for your response.
I've added couple of line notes on c86bb87.
As i see, tests still not covering the case where both parent and child models are proxies.

Kind regards

RamezIssac:
The problem with this pull is the explicit registration of the base model without follow parameter which overrule the admin automatic registration for the proxy model, thus resulting the test fail.
Conversation on #465

PR #484, opened 2016-01-28
Include tests in pypi source tarball.

fladi:
This would allow to run the tests when building packages for a distribution e.g.
Debian.


etianen:
Thanks!

PR #488, opened 2016-02-03
Convert management commands to argparse

### LINKED ISSUE #487: Upgrade management commands to ArgumentParser
Django 1.10 will stop supporting optparse-based management commands.

The question is: will the next django-reversion still support Django < 1.8? If yes, some compat glue code will be needed. If no, things are a bit simpler to update the commands.

claudep:
Fixes #487.


etianen:
Very nice indeed! Appreciate the code cleanup to drop all Django 1.7 hacks too!

claudep:
Thanks for merging so quickly!

Issue #496, opened 2016-03-16
Use of RevisionMiddleware with admin rollback creates empty duplicate revisions


etianen:

etianen:
It might be time to get rid of RevisionMiddleware for good. However, this bug will exist if manual revisions also wrap the admin integration.

acuatoria:
revisions everywhere

letoss:
+1 to fix this

letoss:
I fixed in a fork of mine. But I broke 2 tests and I don't know / don't have the time to fix them. If anyone want to fix them, go ahead! Tell me if you want that I create a PR with the fix.

zvin:
I have been looking into this today and the problem seems to come from VersionAdmin.revisionform_view that calls version.revision.revert(delete=True) that calls DeserializedObject.save that calls Model.save_base that triggers a post_save signal. This signal is received by the middleware and a new revision is created.

I'll make a PR that disconnects the signal handler while temporarily restoring the version.

letoss:
@zvin I think is much simpler than that. If you take a look into the Middleware you'll that is creating a revision even the request method is a GET one. I solve the problem just adding an `if request.method.lower() != 'get'`

etianen:
I realise that this is a bit of a major issue, so I'll weigh in here. I aim
to have a real solution out next week.

Meanwhile, I'd strongly consider not using the RevisionMiddleware, as it
breaks transactional isolation. The best approach is to mark up individual
blocks of code with reversion.create_revision().

letoss:
@etianen I'm not sure how can I do that if I don't have defined the save method in the model. Also, I have to `.create_revision()` in all the save calls, correct?

letoss:
Okay. Using `create_revision` doesn't add the user that made the change.

etianen:
That is also correct, you have to set the user using reversion.set_user().

The RevisionMiddleware is convenient, but when Django deprecated the
TransactionMiddleware they broke transaction isolation of middleware, so
it's not a recommended approach to take at the moment.

etianen:
This is probably fixed in the refactor branch:

https://github.com/etianen/django-reversion/tree/refactor

I'm in the process of refactoring the tests to confirm, but some early testing would be appreciated!

etianen:
Testing seems to confirm that this is now fixed in master branch.

PR #507, opened 2016-04-20
disconnect middleware signal handler when temporarily restoring object to show it in admin form

zvin:
This solves #496.


etianen:
The real solution involves fixing nested revisions. I'd rather get a real fix in early next week then pull in a temporary fix.

Meanwhile, I'm keeping this up as a reference for people who do want a quick fix.

etianen:
This should be fixed in the refactor branch, so closing.

PR #510, opened 2016-04-26
Prevented field.rel deprecation warnings on Django >=1.9

claudep:


etianen:
Nice approach! Thanks for your contribution!

PR #512, opened 2016-04-27
Disable default ordering defined on Model.Meta

ticosax:
a tiny performance optimization.


etianen:
I don't care what people say... I love micro-optimizations. Thanks!

PR #513, opened 2016-04-27
Various Maintenance

ticosax:


etianen:
Thanks for picking this up!

I'm hoping to give reversion some love myself when I've cleared my current deadline, so that's one thing I don't have to do myself now. :)

PR #514, opened 2016-04-29
Return 404 rather then 500 due to ValueError in reversion view methods.

FrankSalad:
Both `recover_view` (at https://github.com/etianen/django-reversion/blob/master/src/reversion/admin.py#L209) and `revision_view` (at https://github.com/etianen/django-reversion/blob/master/src/reversion/admin.py#L219) call `get_object_or_404` which will raise a ValueError if a non-integer is passed to the version_id parameter.


etianen:
A fair point!

PR #515, opened 2016-04-29
One liner refactoring

ticosax:
small chunk of a bigger refactoring I'm trying to reach.


etianen:
Why not indeed!

What are your plans for a refactoring? It would be good to know in advance, as large code dumps can be hard to integrate blind. :)

ticosax:
I tried to add a `GenericRelation` field automatically to the model being versioned. But it appears to be a dead end as it deletes `Version` instances on `Model` deletion (with on_delete behaviour).
I will

PR #517, opened 2016-05-03
Add --no-input so management commands look similar to other Django commands

### LINKED ISSUE #516: Add `--no-input` flag like every other Django command
Currently, Reversion uses `--no-confirmation`, but no one else does that. The expected behavior is `--no-input`.

crccheck:
closes #516 

When Django management commands have a means of skipping confirmation
checks, usually the flag is '--no-input'. The 'deleterevisons' command
uses '--no-confirmation'. Since it functions exactly as you would expect
'--no-input', it makes sense to follow convention. The existing
'--no-confirmation' flag is left deprecated, but there is no schedule
for removal, nor is there a need to remove it any time soon.

Tested by running the test project. I didn't get a deprecation warning to pop up, but I can't tell if that's just me or if I did something wrong.
```
$ cd src/tests

$ ./manage.py migrate

$ python -Walways manage.py deleterevisions
All revisions will be deleted for all models.
Are you sure you want to delete theses revisions? [y|N] y
Deleting revisions...
Done

$ python -Walways manage.py deleterevisions --no-input
All revisions will be deleted for all models.
Deleting revisions...
Done

$ python -Walways manage.py deleterevisions --no-confirmation
/Users/crc/Sync/django-reversion/src/reversion/management/commands/deleterevisions.py:86: DeprecationWarning: --no-confirmation is deprecated, please use --no-input instead
  DeprecationWarning
All revisions will be deleted for all models.
Deleting revisions...
Done

$ ./manage.py deleterevisions --help
...snip...
  --noinput, --no-input
                        Do NOT prompt the user for input of any kind before
                        deleting revisions.
  -c, --no-confirmation
                        Disable the confirmation before deleting revisions
                        (DEPRECATED)
```

I also forced my version of Django to 1.8 got the same results

### DIFF
--- a/src/reversion/management/commands/deleterevisions.py
+++ b/src/reversion/management/commands/deleterevisions.py
@@ -1,6 +1,8 @@
 from __future__ import unicode_literals
 
-import datetime, operator
+import datetime
+import operator
+import warnings
 from functools import reduce
 
 from django.core.management.base import BaseCommand
@@ -57,11 +59,16 @@ Examples:
             action="store_true",
             default=False,
             help="Force the deletion of revisions even if an other app/model is involved")
+        parser.add_argument("--noinput", "--no-input",
+            action="store_false",
+            dest="interactive",
+            default=True,
+            help="Do NOT prompt the user for input of any kind before deleting revisions."),
         parser.add_argument("-c", "--no-confirmation",
             action="store_false",
             dest="confirmation",
             default=True,
-            help="Disable the confirmation before deleting revisions")
+            help="Disable the confirmation before deleting revisions (DEPRECATED)")
         parser.add_argument("-m", "--manager",
             help="Delete revisions only from specified manager. Defaults from all managers.")
         parser.add_argument("--database",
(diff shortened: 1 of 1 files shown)


src/reversion/management/commands/deleterevisions.py
```diff
@@ -71,7 +78,14 @@ def handle(self, *app_labels, **options):
         days = options["days"]
         keep = options["keep"]
         force = options["force"]
-        confirmation = options["confirmation"]
+        interactive = options.get("interactive")
+        if not options.get("confirmation"):
+            interactive = False
+        warnings.warn(
```
etianen: If I read this correctly, the deprecation warning will be emitted irrespective of whether the "confirmation" argument is given. I think we need something like:

``` py
if options["confirmation"] is False:
    interactive = False
    warnings.warn(  # etc...
```

crccheck: oops! I de-dented it to make it warn all the time and forgot to undo before committing.

crccheck:
Ok, I figured out how to get the deprecation warning to show and updated the PR description to use `python -Walways manage.py deleterevisions`. I don't know if end users will ever see the warning since Python doesn't show deprecation warnings by default.

src/reversion/management/commands/deleterevisions.py
```diff
@@ -203,14 +217,14 @@ def handle(self, *app_labels, **options):
...
-        
+
         try:
             revision_query.delete()
         except DatabaseError:
             # may fail on sqlite if the query is too long
             print("Delete failed. Trying again with slower method.")
             for item in revision_query:
                 item.delete()
-
```
crccheck: I have my editor configured to strip trailing whitespace. I tried to `git add -p` just the non-whitespace changes but some got through.

etianen: My editor is set up a lot more strictly than it was when I wrote reversion, so this sort of thing happens all the time! No worries!

etianen:
Appreciate your help here! :)

PR #521, opened 2016-05-15
Change Object_Id to Char and Index

mshannon1123:
- Remove object_id_int
- Combination index for content_type and object_id for performance


etianen:
Aha, I remember discussing this! I'll take a look first thing Monday!

mshannon1123:
Thanks, i'm sure its a rather big change to absorb in current version.  We use oracle in-house so I haven't had an opportunity to try it beat it up on mysql or postgresql (hence the extra commits).  Your great module has been working very wel so thank you.  The only issue was performance, which was partly our fault for deciding not to use integer keys.

As a side note, we also use django-reversion-compare and these changes broke a few things in it.  I'm hoping to make a PR to fix on that project.

src/reversion/functions.py
```diff
@@ -0,0 +1,52 @@
...
+        if output_field_class == 'CharField':
+            template = '%(function)s(%(expressions)s AS %(db_type)s) COLLATE ''utf8_unicode_ci'''
+        return self.as_sql(compiler, connection, template=template)
+
+    def as_postgresql(self, compiler, connection):
+        output_field_class = self._output_field.get_internal_type()
+        if output_field_class in self.override_postgresql_types:
+            self.extra['db_type'] = self.override_postgresql_types[output_field_class]
+        # CAST would be valid too, but the :: shortcut syntax is more readable.
+        return self.as_sql(compiler, connection, template='%(expressions)s::%(db_type)s')
```
etianen: Hi!

Having had a look over your code, this is the only bit that bothers me. It's not really portable between database engines, or with custom fields, which makes me very cautious about including it.

The alternative is to just use in-memory joins for createinitialrevisions and get_deleted. This will scale fine up until a million rows or so, at which point it'll start to eat hundreds of MB of memory.

Your thoughts appreciated!

src/reversion/models.py
```diff
@@ -214,3 +194,6 @@ def __str__(self):
     #Meta
     class Meta:
         app_label = 'reversion'
+        index_together = [
+            ["content_type", "object_id"],
```
etianen: I think this needs to be ["object_id", "content_type"], since that'll mean the index can be used just for "object_id", for queries that don't need the content_type.

mshannon1123: The only downside to that is then we would need an index on content_type for FK validation.

All your internal logic always specifies the content_type, but I understand others might not expect that situation.  (Though it would be dangerous to do that lookup and not specify the content_type, since the object_id is not guaranteed to be unique across content_types.)

If you still want me to switch it and re-add the FK, it's fine with me.

etianen: Because content_type is a ForeignKey, it'll have an index anyway, so I think all bases are covered.

src/reversion/functions.py
etianen: Hold fire for the time being, I've almost got a solution that supports common database setups for in-database joins.

mshannon1123: Hah! Didn't see your comment till now, I pushed up a commit with it changed to in-memory.  

I did add a unit test to check various primary key types, so that might be useful to retain.

src/reversion/models.py
mshannon1123: Changed

etianen:
So, I'd appreciate your thoughts on the approach in #527.

Closing this for now, as I think the in-database support for joins supersedes this approach. Very much appreciate your work on this up to now, and will mention you as a contributor to this feature in the next release notes. :)

Issue #524, opened 2016-05-25
Does django-reversions support Oauth2 token?


bmarika:
First of all thanks for the awesome lib.

I am using the django reversion middle ware in my application and I noted that at some point all the ``version[i].revision.user`` was None. Then it hit me that I had switched to token based authentication when I saw the line https://github.com/etianen/django-reversion/blob/master/src/reversion/middleware.py#L33
I did a quick fix by removing the session related checks since on searching the repo, it was only being used at that point. Everything worked well.

What are the implications of removing the session related checks? Can we remove them so as to work seamlessly with token authentication?

PR #525, opened 2016-05-25
Remove session checks to allow for a token authenticated user to be set.

bmarika:


bmarika:
Resolves https://github.com/etianen/django-reversion/issues/524

src/reversion/middleware.py
```diff
@@ -29,9 +29,7 @@ def _close_revision(self, request):
 
     def process_response(self, request, response):
         """Closes the revision."""
-        # look to see if the session has been accessed before looking for user to stop Vary: Cookie
-        if hasattr(request, 'session') and request.session.accessed \
-                and hasattr(request, "user") and request.user is not None and request.user.is_authenticated() \
+        if hasattr(request, "user") and request.user is not None and request.user.is_authenticated() \
```
etianen: Are we absolutely certain that this won't result in the session being accessed?

I was actually wondering about making the check simply inspect request.method, and set the user if it's a POST, PATCH or PUT request. It's simpler, and mutating methods shouldn't suffer from losing the cache.

bmarika:
@etianen Am also not sure on whether the session will be accessed.  Let me make the change and check for the request.method

etianen:
Thanks! Just merged!

PR #528, opened 2016-05-31
Additional Primary Key Tests

mshannon1123:
Pulled over the primary key tests if you would like them.


etianen:
Thanks! I'm actually in the process of performing a major cleanup of the tests in the refactor branch, so these probably won't exist for very long, but they're a good verification of the current codebase.

mshannon1123:
haha oh well!   I saw you implemented tox, it worked well locally, might want to note that users will need to install postgresql and mysql to run it.

etianen:
Good point!

PR #532, opened 2016-06-08
many api has moved from reversion to reversion.revisions

whitekid:


etianen:
Not in the master branch, which is what the current docs are referring to. :)

However, I can see this is going to cause confusion, so I've stuck the docs for 1.10.3 up as the default version. It should redirect you to the 1.10.3 docs when you visit the site now.

https://django-reversion.readthedocs.io

Issue #541, opened 2016-06-16
Only run de_dupe_version_table (migration 3) if setting the field to unique fails


blueyed:
`de_dupe_version_table` is a very heavy query, and might not be necessary, if the UNIQUE from migration 4 would work right away.

It would be great if it could be skipped and/or optimized in general.

```sql
EXPLAIN SELECT "reversion_version"."id" FROM "reversion_version" WHERE NOT ("reversion_version"."id" IN (SELECT MAX(U0."id") AS "max_pk" FROM "reversion_version" U0 GROUP BY U0."revision_id", U0."content_type_id", U0."object_id")) ORDER BY "reversion_version"."id" DESC
```

```
Index Only Scan using reversion_version_pkey on reversion_version  (cost=242852.85..10696466083.66 rows=566583 width=4)"
  Filter: (NOT (SubPlan 1))"
  SubPlan 1"
    ->  Materialize  (cost=242852.42..261204.41 rows=210561 width=16)"
          ->  GroupAggregate  (cost=242852.42..259122.61 rows=210561 width=16)"
                Group Key: u0.revision_id, u0.content_type_id, u0.object_id"
                ->  Sort  (cost=242852.42..245685.34 rows=1133166 width=16)"
                      Sort Key: u0.revision_id, u0.content_type_id, u0.object_id"
                      ->  Seq Scan on reversion_version u0  (cost=0.00..109532.66 rows=1133166 width=16)"
```

PR #542, opened 2016-06-16
do not do heavy queries if no objects would be deleted anyway

mathiasertl:
As mentioned in #541, de_dupe_version_table in migration 0003 is very heavy. At least in my case, it turned out that the query wasn't necessary at all:

```
>>> from reversion.models import Version
>>> from django.db import models
>>> keep_version_ids = Version.objects.order_by().values_list(
        # Group by the unique constraint we intend to enforce.
        "revision_id",
        "content_type_id",
        "object_id",
    ).annotate(
        # Add in the most recent id for each duplicate row.
        max_pk=models.Max("pk"),
    ).values_list("max_pk", flat=True)
>>> keep_version_ids.count()
806866
>>> Version.objects.all().count()
806866
```

The subsequent statement is what takes so incredibly long, so this pull request would be a significant speedup in this case (I don't know how likely this is, but at least it was true for me)

### DIFF
--- a/reversion/migrations/0003_auto_20160601_1600.py
+++ b/reversion/migrations/0003_auto_20160601_1600.py
@@ -21,11 +21,9 @@ def de_dupe_version_table(apps, schema_editor):
         # Add in the most recent id for each duplicate row.
         max_pk=models.Max("pk"),
     ).values_list("max_pk", flat=True)
-
     # Do not do anything if we're keeping all ids anyway.
     if keep_version_ids.count() == Version.objects.all().count():
         return
-
     # Delete all duplicate versions. Can't do this as a delete with subquery because MySQL doesn't like running a
     # subquery on the table being updated/deleted.
     delete_version_ids = list(Version.objects.exclude(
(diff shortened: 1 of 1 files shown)


reversion/migrations/0003_auto_20160601_1600.py
```diff
@@ -21,6 +21,11 @@ def de_dupe_version_table(apps, schema_editor):
         # Add in the most recent id for each duplicate row.
         max_pk=models.Max("pk"),
     ).values_list("max_pk", flat=True)
+
+    # Do not do anything if we're keeping all ids anyway.
+    if keep_version_ids.count() == Version.objects.all().count():
+        return
+
```
etianen: Looks like a very sensible optimization. Can I be picky and ask for the whitespace around the newlines to be removed to match the rest of the style? Then I'll merge right away.

mathiasertl:
done! my local migration has been running for over an hour now -.-

etianen:
Nice. It's annoying that the de-duping needed to be done in the first place, as that unique constraint should have been enforced at the software level since the beginning of the project.

Just goes to show the usefulness of db-level constraints.

New users just get the squashed migration, anyway, so this is a temporary problem.

blueyed:
Thanks!

PR #552, opened 2016-07-08
Feature : custom User model

GuillaumeCisco:


etianen:
Thanks for the pull request!

Unfortunately, setting this to anything other than AUTH_USER_MODEL will break the admin integration, revision middleware and revision view helpers. I don't think it belongs in django-reversion,

If you want to associate a non-admin user model with a revision, consider using the add_meta() feature:

https://django-reversion.readthedocs.io/en/stable/api.html#revision-api

PR #561, opened 2016-07-25
Missing database index on object_id!

aidanlister:
There's a missing index on object_id, this is creating a sequential scan instead of an index lookup when grabbing all reversions for an object: http://share.aidanlister.com/gmPY

### DIFF
--- a/reversion/models.py
+++ b/reversion/models.py
@@ -155,6 +155,7 @@ class Version(models.Model):
     object_id = models.CharField(
         max_length=191,
         help_text="Primary key of the model under version control.",
+        db_index=True,
     )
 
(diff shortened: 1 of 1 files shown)


etianen:
How are you generating that query? The built-in revision lookup API methods all include a filter on the `db` column too, allowing this unique index to kick in:

https://github.com/etianen/django-reversion/blob/master/reversion/models.py#L267

etianen:
Actually, that query plan is from an old version of django-reversion. The object_id_int column no longer exists!

aidanlister:
Oh, sorry! Yes we're using an old version.

etianen:
Phew!

Issue #559, opened 2016-07-20
squashed migration broke django-reversion upgrade.


xuchen81:
Issue:
Django version: 1.9.7
django-reversion version: 2.0.5
I was using django-reversion 1.8.0, and there is no "db" columns in reversion_version table.
Now I am upgrading to version 2.0.5, it adds the column "db" in migration 3, but since there is also a squashed migration file there, when I run python manage.py migrate --fake-initial, it will fake the squashed migration and ignore all the other migrations in the folder, and the migration 2,3,4 will never be ran, so "db" column will not be added to the existing reversion_version table.

Suggested solution:
Remove squashed migration.

Thanks.

etianen:
Why are you faking the initial migration?

This isn't an issue with django-reversion. It *might* be an issue with Django's migration framework.

alukach:
We are also getting this error.  I believe that the issue is based on the fact that in `reversion/migrations/0003_auto_20160601_1600.py`, the code is both a) adding the `db` column (lines [94-98](https://github.com/etianen/django-reversion/blob/master/reversion/migrations/0003_auto_20160601_1600.py#L94-L98)); and then b) using the `db` column (line [66](https://github.com/etianen/django-reversion/blob/master/reversion/migrations/0003_auto_20160601_1600.py#L66)).

From the [RunPython section of the Django Migration docs](https://docs.djangoproject.com/en/1.9/ref/migration-operations/#django.db.migrations.operations.RunPython), it states:

> On databases that do support DDL transactions (SQLite and PostgreSQL), **RunPython** operations do not have any transactions automatically added besides the transactions created for each migration (the **atomic** parameter has no effect on these databases). Thus, on PostgreSQL, for example, you should avoid combining schema changes and **RunPython** operations in the same migration or you may hit errors like **OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events**.

Basically, you can't create a column and then use it in the same migration with a database that supports DDL.  My guess is that @xuchen81 is using either SQLite or PostgreSQL and @etianen is using MySQL or Oracle.

I'm unsure of where we should go from here.  Possibly break out the `RunPython` code into a separate migration named `0003_set_version_db` and add that as a dependency for `0004_auto_20160611_1202.py`?

etianen:
[...]

PR #563, opened 2016-07-27
Use db-aware queries in migration

alukach:
In relation to #559.

I believe using the `.using()` method on the object manager is best-practice for Django migrations.  I would like to see if @xuchen81 can verify that this resolves his issues before merging this.


etianen:
Thanks!

The travis build failed due to the linter randomly exploding, which I don't really understand. I'll deal with that separately.

PR #565, opened 2016-07-31
Added django app config module.

Alkalit:
See: https://docs.djangoproject.com/en/1.9/ref/applications/

### DIFF
--- a/reversion/__init__.py
+++ b/reversion/__init__.py
@@ -36,4 +36,6 @@ else:
         get_registered_models,
     )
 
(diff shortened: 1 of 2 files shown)


etianen:
Thanks for looking into this.

Could you explain the benefit of adding this to django-reversion? My understanding was that an app config module was entirely optional.

Alkalit:
The reason for this PR is for I want to make translation of this app name in django-admin interface. I will add translate on russian in next PR.

etianen:
I think that you can do this without requiring any changes to django-reversion. You can place the AppConfig anywhere in your codebase.

Issue #564, opened 2016-07-28
Add distinguishing context variable to the recover/revision view contexts


kezabelle:
For reasons which are somewhat convoluted, I have a need to know within the context which change form may be being rendered, but I cannot see a way of doing so without overriding the appropriate views to pass along extra context at every modeladmin necessary.

Would it be feasible to update the context given to `_reversion_revisionform_view` to include unique keys saying who the original caller is, i.e:

```
context.update(extra_context or {})
context.update(revision_view=True)  # or recover_view=True
```

etianen:
You can do this yourself by subclassing VersionAdmin, and overriding
revision_view and recover_view, adding the appropriate keys to
extra_context. That should probably cover your use-case.

kezabelle:
Hiya etianen. Thanks for taking the time to respond.
Yeah, that's actually what I was trying to avoid doing (and what I was badly alluding to with: `overriding the appropriate views to pass along extra context at every modeladmin necessary.`). I figured that ModelAdmin itself already sets the precedent for distinguishing context vars with `add` and `change` and thought perhaps you'd be amenable to it. 

I have managed a workaround without touching extant modeladmins, but its grotty because it involve slicing the start of the context `title` and comparing it to a `ugettext` string of "Revert" or "Recover" so I'm not sure how well it'd hold up in a i18n scenario, but it works well enough for my purposes for now.

I'm happy for you to close this down if you don't want to burden yourself with supporting potentially unnecessary code :)

etianen:
I had a quick scan of the Django source code, and couldn't see where
passing along "add" and "change" is done. If there's a precedent, I'm happy
to continue it.

Otherwise, you should only need to subclass VersionAdmin once, then use
that subclass in all your other ModelAdmins.

kezabelle:
`add` is defined in `changeform_view` [here](https://github.com/django/django/blob/master/django/contrib/admin/options.py#L1421) and the distinguishers (for *state*, rather than *view*, but principally accomplishes the same) are added to the context in `render_change_form` [here](https://github.com/django/django/blob/master/django/contrib/admin/options.py#L1040-L1041).

[...]

PR #566, opened 2016-08-02
added `revert` and `recover` context variables to mirror Django's

kezabelle:
As per conversation in #564 (where I misnamed the potential variables, but we ultimately settled on a pair), this adds new context variables to a VersionAdmin when rendering the changeform view, which allow the template or middleware to establish which view is being rendered. Their absence indicates it was not the view in question.


etianen:
Thanks!

Meanwhile, something very odd is happening with travis CI, and the build is breaking for really odd reasons all of a sudden...

PR #569, opened 2016-08-15
Remove duplicated API doc

jschneier:
I noticed that this is included twice in a row, thanks for the project!


etianen:
Sorry I took so long to merge, I was away on holiday. :)

Thanks for the fix!

Issue #571, opened 2016-08-24
`post_revision_commit` is not perfectly replaceable by `post_save` of `Revision`


carlosxl:
Saw the `post_revision_commit` signal is removed in 2.0 release. The suggested way is to replace the existing `post_revision_commit` with `post_save` of `Revision`.

However, there is a difference that with `post_revision_commit`, the `versions` objects have already been saved to db whereas with the `post_save` of `Revision` they are not.

Got a hard time migrating from 1.9 to 2.0 because of this issue. Any suggestion?

etianen:
Use the post_commit hook in transactions? That's generally a much better
place for post_save behaviour when the side-effects do not involve
modifying the database.

Use the post_save signal of Version?

I see your point though, and would not be adverse to putting the signal
back in if there's no alternative.

lxiang-visualsquares:
Thanks for the suggestions!

The post_commit or on_commit hook feels like to have a more generic purpose.

On the other hand, we do have a use case in which multiple Versions are created inside a single Revision and we want to wait until all the versions are saved.

In my case, I think the `post_revision_commit` is still valuable if the user want to perform some tasks that assume the `revision` and all its related `versions` have been saved to db.

carlosxl:
Thanks for the suggestions!

The post_commit or on_commit hook feels like to have a more generic purpose.

On the other hand, we do have a use case in which multiple Versions are created inside a single Revision and we want to wait until all the versions are saved.

In my case, I think the post_revision_commit is still valuable if the user want to perform some tasks that assume the revision and all its related versions have been saved to db.

PR #572, opened 2016-08-25
Reintroduce pre_revision_commit and post_revision_commit signals

carlosxl:
Rationale: https://github.com/etianen/django-reversion/issues/571
Related commit: 09ce3764c1404747ebb90379cd79e193f3f58164

### DIFF
--- a/docs/_include/signal-args.rst
+++ b/docs/_include/signal-args.rst
@@ -1,0 +1,8 @@
+``sender``
+    The ``reversion.create_revision`` object.
+
+``revision``
+    The :ref:`Revision` model.
+
+``versions``
+    The :ref:`Version` models in the revision.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -70,3 +70,4 @@ Usage
    views
    middleware
    errors
+   signals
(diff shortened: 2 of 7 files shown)


carlosxl:
@etianen not sure
1. if the signal signature aligns with your design; and
2. if this is the right strategy to test the signals.

Please take a look, thanks!

etianen:
This looks perfect!

1. Sure does!
2. I've not used mock much, but I like how those tests look. :)
3. I believe that the signals are dispatched inside the transaction block, which is good, as it contains their database side-effects inside the same transaction.

If you could also add a `signals.rst` to the docs, and link it in, that would finish this perfectly.

(Unfortunately, having spent a long time updating the docs recently, I've since come to realize that there are much better ways of formatting them using python hooks built into sphinx. Please follow the existing style for now, even if it's a bit odd!)

carlosxl:
Took a shortcut to copy most of it from the original doc. See if it is consistent with the existing style :)

etianen:
Lovely job, thanks!

carlosxl:
Great project here! Thanks!

Issue #573, opened 2016-08-25
Django 1.10 incompatibility


KhasanovBI:
Hey,

I just upgraded Django to 1.10, and here's the error:
Internal Server Error: /api/login/
```
Traceback (most recent call last):
  File "/Users/b.khasanov/Envs/intranet/lib/python2.7/site-packages/django/core/handlers/exception.py", line 39, in inner
    response = get_response(request)
  File "/Users/b.khasanov/Envs/intranet/lib/python2.7/site-packages/django/core/handlers/base.py", line 244, in _legacy_get_response
    response = middleware_method(request)
  File "/Users/b.khasanov/Envs/intranet/lib/python2.7/site-packages/reversion/middleware.py", line 24, in process_request
    _set_user_from_request(request)
  File "/Users/b.khasanov/Envs/intranet/lib/python2.7/site-packages/reversion/views.py", line 10, in _set_user_from_request
    if hasattr(request, "user") and request.user.is_authenticated():
TypeError: 'bool' object is not callable
```
In django1.10 is_authenticated is property.

etianen:
Are you using a custom user model? The docs seem to indicate that is_authenticated will continue to work as a method for the time being: https://docs.djangoproject.com/en/1.10/releases/1.10/#using-user-is-authenticated-and-user-is-anonymous-as-methods

KhasanovBI:
Yes, you right, thanks. But for future I think it makes sense to make the compatibility for the future. If you don't mind, I'll do pull request.

GuillaumeCisco:
I have an issue too, but really weird :/
Don't understand where it comes from.
When I try to create an object from django rest framework with their Browsable API, I get:
```
 ... object has no attribute '_deferred'
```

etianen:
@GuillaumeCisco - I don't think that this is a django-reversion error. I suspect that you're using an incompatible version of django-rest-framework and django.

GuillaumeCisco:
@etianen Thank you for your time.
Just retested it with django 1.10.2 and djangorestframework 3.4.7.
The issue was still here.
The issue was with the django-model-utils dependency version 2.5, just upgraded to 2.6 and it works fine.
Sorry for the inconvenience.

etianen:
No probs! :)

PR #574, opened 2016-08-25
Add `is_authenticated` compatibility django 1.10+

KhasanovBI:


KhasanovBI:
refs #573

etianen:
Thanks!

I think we'd be better off using capability testing, rather than version checking here. Something like isinstance(user.is_authenticated, bool)

KhasanovBI:
But in django 1.10 `is_authenticated` can be a property or attribute.

etianen:
I meant like:

``` py
if isinstance(user.is_authenticated, bool):
    return user.is_authenticated
return user.is_authenticated()
```

That should work for the case of CallableBool and bool.

KhasanovBI:
But that's not work in case of CallableBool
image
So in django with version less than 1.10 I can't import CallableBool to check isinstance of it.

etianen:
Damn, you're right. 

I hate using version number checks, but in this case I can't see a better alternative.

PR #575, opened 2016-09-02
Fix docs for `Version.objects.get_for_object_reference`

zsiciarz:
This fixes a copy&paste mistake in API docs.


etianen:
Good spot, ta!

PR #580, opened 2016-09-13
Fixed revision message format on Django 1.10

### LINKED ISSUE #579: Revision comment format on Django 1.10
On Django 1.10, the LogEntry `change_message` is often a list or a JSON structure. It would be better to retrieve the real message through `LogEntry.get_change_message()`, so as it is saved as a string in the revision comment. I'll see if I can come up with a test for that...

claudep:
Fixes #579.

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -1,9 +1,11 @@
 from __future__ import unicode_literals
+import json
 from contextlib import contextmanager
 from django.db import models, transaction, connection
 from django.conf.urls import url
 from django.contrib import admin, messages
 from django.contrib.admin import options
+from django.contrib.admin.models import LogEntry
 from django.contrib.admin.utils import unquote, quote
 try:
     from django.contrib.contenttypes.admin import GenericInlineModelAdmin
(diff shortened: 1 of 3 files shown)


claudep:
Unfortunately, I'm not understanding test failures yet and struggling to see the relation with my PR.

claudep:
Order of tests is different on Python 2 and 3. If the test_app is registered when the admin.site.get_urls is called (once for all the test suite), then the tests run fine. Conversely, if test_app is not registered with the admin when get_urls is called, several tests are failing during URL resolution.
This is completely unrelated to this patch.

etianen:
Thanks! I've applied a fix to master that will hopefully solve the test ordering issue too. :)

It feels like time for a new release. I'll let things stabilise for a few days, then push something out.

etianen:
So tell me, how are these "pull request reviews" at all different from comments? :P

claudep:
Thanks!

RamezIssac:
Maybe that is little bit late,, But may i ask why are you saving the change_message text result instead of the change_message Json ?!

Thing is 
1- This violates the `no translated text in database` rule 
2- more importantly This will not yield correct translation if language changed

claudep:
It would make sense to keep JSON in comments, even if it could break compatibility for people accessing `Revision.comment` directly.
Please open a separate issue for that.

etianen:
Except... reversion comments are not just used for admin change messages.
They're used for any arbitrary user-supplied change message.

What about people who want to write arbitrary text in the reversion comment?

claudep:
If we mimic the contrib.admin `LogEntry` model behavior, in the comment getter method, we can try first to JSON-decode the message, then fallback to plain text if that doesn't succeed.

etianen:
I guess that makes sense. It would be a breaking change, however, since
existing users would have to use the new getter method everywhere.

PR #589, opened 2016-09-29
Fix for checking db savepoints

aloskutov-ki11:
Fix for checking db savepoints

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 import json
 from contextlib import contextmanager
-from django.db import models, transaction, connection
+from django.db import models, transaction, connection, DatabaseError
 from django.conf.urls import url
 from django.contrib import admin, messages
 from django.contrib.admin import options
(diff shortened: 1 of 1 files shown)


etianen:
Thanks for the pull request.

What problem is it fixing?

aloskutov-ki11:
All my projects are encapsulated in docker containers and Dockerfile for container with django has string `RUN python ./manage.py collectstatic --noinput`. If you run `docker build server` container has no access to database and build breaks.

etianen:
I don't think that accessing connection.features opens a database connection. Is it that the docker container doesn't contain the database settings, which is causing an ImproperlyConfigured exception to be raised?

If so, surely the fix is to provide the database settings to the container?

Alternatively, we could move the check to the recover/revert view to delay the check until runtime.

aloskutov-ki11:
Yes you are right. All database settings I put in a file with environment variables and container building process knows nothing about them. This .env file is connected to the container only when it's lainched

PR #594, opened 2016-10-13
Release as a universal wheel

adamchainz:
By releasing as a [Python wheel](http://pythonwheels.com/) as well as a source distribution, you can speed up end user’s installs. After merging this command, to release you just need to run `python setup.py clean sdist bdist_wheel upload`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal = 1


adamchainz:
I'm actually unsure how this integrates with the `babel.messages` stuff you're doing in `setup.py` though, gonna close for now

etianen:
Actually, I'd like this open.

I don't actually know that the babel.messages stuff is either. It was a request made a long time ago by another user.

etianen:
I'd not bothered doing this before, because it's a small pure python package. Still, if it can speed up installs, there seems no reason not to do it. :)

adamchainz:
What I meant was I'm not entirely sure you can make wheels at all because of the conditional logic in `setup.py` around `babel.messages`, but looking again you should be able to because it only affects people invoking `setup.py` to update the message catalogs afaict.

etianen:
I'll try it out for the next release, anyway. I'm looking to push it out at the start of next week.

Thanks for the suggestion and pull request!

etianen:
Just release 2.0.7 as a universal wheel!

https://pypi.python.org/pypi/django-reversion/2.0.7

adamchainz:
Yay!

PR #597, opened 2016-11-03
Add changelog note on models ordering change done in 2.0.0

jaywink:
Noticed the ordering of the `Revision` and `Version` models was reversed in 2.0, [in this commit](https://github.com/etianen/django-reversion/commit/8a3a6c369fd4d16253884dd31c98969065d1e26e#diff-848d9ceb3ff77cf47a950aa6c22852daR99). This caused a little confusion when updating, since `.last()` was now `.first()`.

A changelog entry on this might help future upgraders to check their code.

Thanks for your work on the library.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -248,6 +248,10 @@ Helpers
 
     The old implementation is available for reference from the `previous release <https://github.com/etianen/django-reversion/blob/release-1.10.2/src/reversion/helpers.py>`_.
 
+Models
+^^^^^^
+
+* **Breaking:** Ordering of ``-pk`` added to models ``Revision`` and ``Version``. Previous was the default ``pk``.
 
 1.10.2 - 18/04/2016
 -------------------


etianen:
Interesting. So the "-pk" ordering was applied if you used the reversion helpers of get_for_object() or get_for_object_reference(), but wasn't applied as the default ordering of the models? And I guess you were previously performing queries on the models themselves, rather than using the helpers?

jaywink:
A model is using a `GenericRelation` like this:

    versions = GenericRelation("reversion.Version")

Since the `Version` table previously was ordered by `pk`, the new `-pk` added in the referenced commit turned results upside down in ordering. Just though it might be helpful to have that noted in the changelog, just in case someone else uses the models directly :)

etianen:
Definitely. Most people were simply using the helpers, but I added the default ordering in to ensure consistency.

Didn't think I'd break anybody's code! :O

Thanks for the contribution. :)

PR #601, opened 2016-11-10
Update recover_list.html

w4rri0r3k:
When deletion.pk is larger than 1k  in some localizations settings Django adds space between thousnads (for example: "1 000") and cause bug in url (creates inaccesible url)


etianen:
What strange behaviour to include by default!

Thanks for the fix. 😄

PR #611, opened 2016-12-30
Speedup 0003 migration

mkurek:
Migration 0003 could be really slow on DBs with a lot of revisions/versions. This PR speed it up by:
* using content type id instead of app label and model name, thus eliminating join in update
* since multi-db is used by relatively small amount of Django apps, there is special path for single db (default) apps to not filter rows, but update all of them


etianen:
Nice! Thanks.

PR #616, opened 2017-01-11
Details on deleting revisions

orlra:
More details on deleting revisions. Most typical usage.
3 commands I've found most useful with details what will happen.
Code is nice and allows stacking of --keep and --days, but nothing about that in docs.


etianen:
Very nice. One thing - could you put the comments for the commands above the related command, rather than below? I think that would be more understandable for people.

etianen:
Thanks!

PR #625, opened 2017-02-26
Issue 624 optional atomic transactions

ernestoferro:
A few notes:

- I'm not super happy with revision._dummy_context, but it's the best compromise I could think of. See the second commit for an explanation.
- I updated the docs but didn't build them. I'm pretty sure it's OK, but just in case double check.
- I added tests for the api, view, mixin and middleware. The last three rely on new views which return True if an atomic transaction was active in that view. It's not great because you are trusting that the test view itself is OK, but since the transaction is done by the time I get the response I couldn't find a nice way of testing this. Mocking might be an alternative here.
- This test fails in master testDeleteRevisionsDbMySql (test_app.tests.test_commands.DeleteRevisionsDbTest) and I didn't fix it.

### DIFF
--- a/docs/_include/create-revision-args.rst
+++ b/docs/_include/create-revision-args.rst
@@ -3,3 +3,6 @@
 
 ``using``
     .. include:: /_include/create-revision-using.rst
+
+``atomic``
+    .. include:: /_include/create-revision-atomic.rst
--- a/docs/_include/create-revision-atomic.rst
+++ b/docs/_include/create-revision-atomic.rst
@@ -1,0 +1,1 @@
+If ``True``, the revision block will be wrapped in a ``transaction.atomic()``.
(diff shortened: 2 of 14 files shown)


ernestoferro:
I didn't know about ExitStack. Thanks for sharing!

And regarding the tests, since it's not my project I tend to lean towards
more tests. But on the other hand, if I forget to pass the atomic parameter
to reversion.create_revision then the middleware test will catch that.

Anyway, I'm happy you liked the changes.

Regards,
Ernesto.

etianen:
Lovely jubbly! Thanks for the contribution! This will be included in the next release, sometime soon. :)

ernestoferro:
Excellent. Thanks Dave.

amureki:
@ernestoferro thank you for important change!

@etianen can we, please, make a release with it? :)

etianen:
v2.0.9 is currently being built, and will be on PyPi in about 30 minutes.

amureki:
@etianen thank you much! I appreciate that! 🎉

Issue #630, opened 2017-03-16
Does reversion 1.10.x support Django 1.10?


guettli:
We are unsure if reversion 1.10.x supports Django 1.10?

Or should we upgrade to reversion 2.x?

According to this page, reversion 1.10.x should support Django 1.10: http://django-reversion.readthedocs.io/en/latest/django-versions.html

etianen:
I'm afraid you're reading that table wrong. If you have Django 1.8+, you need reversion 2.00+.

etianen:
(I can see how the confusion might arise 😄 )

TZanke:
| Django version        | Reversion release           |
| ------------- |:-------------:|
| 1.8-current      | 2.0.0 |
| 1.7      | 1.10.0      | 
| 1.6 | 1.8.5      |

etianen:
Yes, that would make more sense. I'll happily take a pull request to fix it.

guettli:
I created a pull request. https://github.com/etianen/django-reversion/pull/631

PR #631, opened 2017-03-21
More precice supported version matrix

guettli:
See #630


etianen:
Thanks! I'm currently on holiday, being lazy, so thanks for sorting this.

guettli:
You should not be here on github, if you are on holidays :-) have a nice time.

PR #637, opened 2017-04-08
Update changelog with additional breaking change in API

mtully-blitz:
Was updating from 1.10.1 to 2.0.8 and found this additional breaking change. 

Here's when it was changed: https://github.com/etianen/django-reversion/commit/85ae48c3266d726e38d45778dca6be26f0465647#diff-848d9ceb3ff77cf47a950aa6c22852daL190


etianen:
Ah, I didn't realize that was ever part of the public API, hence moving it to a protected attribute.

Still, good spot.

PR #649, opened 2017-06-22
Updated '_set_user_from_request' to make sure that request user is not empty

### LINKED ISSUE #648: Middleware crashes when `request.user` is set`None`
Hiya,

I just wanted to mention that `Setting revision.user in process_response for middleware (@etianen).` feature from release `2.0.8` broke our application because we had the `request.user` set to `None` (not explicitly, but that's how our `request` looked with a Django REST Framework API + OAuth scope based authentication) and https://github.com/etianen/django-reversion/blob/master/reversion/views.py#L18 crashed with `AttributeError: 'NoneType' object has no attribute 'is_authenticated'`.

A basic fix (well, except for downgrading which was what we did : ) would be just changing `hasattr(request, "user")` to `getattr(request, "user", None)`.

pawelad:
Currently this code fails if `request.user == None` with `AttributeError: 'NoneType' object has no attribute 'is_authenticated'`.

It could be worth adding a test which I can do in the evening.

Fixes #648

### DIFF
--- a/reversion/views.py
+++ b/reversion/views.py
@@ -15,7 +15,7 @@ def _request_creates_revision(request):
 
 
 def _set_user_from_request(request):
-    if hasattr(request, "user") and is_authenticated(request.user) and get_user() is None:
+    if getattr(request, "user", None) and is_authenticated(request.user) and get_user() is None:
         set_user(request.user)
 
 


etianen:
Thanks for the pull request! No need for a test, given the simple functionality. Merged!

pawelad:
@etianen well, the test would be to make sure that it doesn't regress in the future and break our builds again : )

etianen:
I understand, but when testing a tiny piece of functionality, you're got to
consider the overhead of maintaining the test code itself.

I'd certainly accept a pull request for a small, simple regression test.

(Sometimes I receive a few hundred lines of tests for a single, simple
change, and test code is still code to be maintained by me in the long term)

PR #655, opened 2017-07-24
One too many S's in MyISAM

danielquinn:


etianen:
Things to hate about MyISAM tables:

1. They're not transactional.
2. I can't spell them.

PR #662, opened 2017-08-22
Fix issue with admin and special characters in primary keys

### LINKED ISSUE #661: Admin broken when using primary keys with underscores
If you have a model with a text based primary key with any of the special characters which django admin quotes, the history list is unable to be browsed. Once that is fixed, the changes are unable to be reverted due to a similar issue.

It's a pretty simple fix, pull request incoming.

achidlow:
Issue is as described in #661. Fix is pretty straightforward, just needed to quote/unquote in the right places.

Also adds unit tests that demonstrate issue and verify the fix.

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -184,7 +184,7 @@ class VersionAdmin(admin.ModelAdmin):
                 version.revision.revert(delete=True)
                 # Run the normal changeform view.
                 with self.create_revision(request):
-                    response = self.changeform_view(request, version.object_id, request.path, extra_context)
+                    response = self.changeform_view(request, quote(version.object_id), request.path, extra_context)
                     # Decide on whether the keep the changes.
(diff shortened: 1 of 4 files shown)


achidlow:
I've reduced it down to a single unit test. I think it would be good to have to check for regressions, as it's an easy thing to forget when making changes. If you'd prefer not to have any unit tests at all let me know and I'll remove all together.

Not sure why travis is failing. It doesn't look like it's caused by the addition of the unit test? At least, not in an obvious way. Any ideas?

etianen:
Cool, thanks. I think this is a good compromise on tests. :)

I'll see if I can work out why the tests are breaking on Python 2.7...

etianen:
For some reason, the tests stopped breaking. Odd. :S

Thanks for your time on this!

achidlow:
No worries, thanks for the super quick merge!

PR #665, opened 2017-09-18
avoid comparing Django version each time is_authenticated is called

fpischedda:
According to some simple test with timeit, this change can save 0.06 to 0.08 seconds on each call to is_authenticated

### DIFF
--- a/reversion/compat.py
+++ b/reversion/compat.py
@@ -24,4 +24,4 @@ def _choose_is_authenticated():
     return django_1_10_and_later
 
 
-is_authenticated = _choose_is_authenticated
+is_authenticated = _choose_is_authenticated()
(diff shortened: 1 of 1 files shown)


etianen:
I'm suprised that a simple `<` comparison to a tuple costs 70ms!

Still, this doesn't really feel like a critical path for any real application. A given HTTP request will involve database connections, HTTP parsing, plus several thousand such trivial variable comparisons. I'm afraid this feels like a pretty trivial micro-optimisation.

fpischedda:
Yep is not that big win, the thing I thought of is "why do this comparison when it is not needed?"

etianen:
I guess because it saves 8 lines of code, and made sense at the time. :man_shrugging: 

I don't really subject little bits of functionality like this to such rigorous analysis!

PR #667, opened 2017-09-24
Add tests for Django master, allowing failures

claudep:


claudep:
#668 should be solved first

etianen:
Thanks!

claudep:
Thanks, but sorry, I was not yet certain it worked, it would have been better to wait for a successful test run first :-)
I think we also need to allow the master failures at tox level (probably with `ignore_outcome`)

claudep:
You might try to add:
```
ignore_outcome =
    djangomaster: True
```
under `[testenv]`

etianen:
Thanks, that fixed it.

What with pyenv, too and travis, there's a lovely mess of configuration at
play here, in an environment you can only access with a git push. *Sigh...*

claudep:
Sure, but you can also test with pull requests before committing to the main master branch (and force push to pull requests until the tests pass).

PR #669, opened 2017-09-25
Fixed tests with Django 2.0

claudep:


claudep:
Not ready, still two errors to fix.

claudep:
Now ready.

claudep:
One change affects the main code, so a new release will be needed to support Django 2.0.

etianen:
Damn fine work!

etianen:
While Django 2.0 is in development, I think it's fine to require installing django-reversion from the GitHub repo.

claudep:
Sure no rush. However, Django 2.0 is feature frozen now, so if we can have a release during the next two months, it encourages people to test the new Django before its .0 release.

PR #679, opened 2017-11-16
adding the ability to include or exclude models while using the revis…

yazanpsut:
…ion middleware.


yazanpsut:
@yorkedork @bartek @charettes @blueyed any comments about this pull request?

etianen:
Thanks for contributing!

I'm confused by this. The proposed setting seems to exclude certain models from version control altogether, not just with RevisionMiddleware.

yazanpsut:
@etianen well... It seemed legit when you are using the middleware because it controls all models, but your point is correct where when ever you use it, you can include or exclude models at all time.

etianen:
You can simply exclude models by not registering them with reversion, surely?

yazanpsut:
@etianen but if you are using the middle ware, can you do that?

etianen:
The middleware will only version models that have been registered with
reversion, either via reversion.register() or via the admin integration.

PR #681, opened 2017-11-27
Add __repr__ methods for Version and Revision

blueyed:
This makes e.g. `list(Version.objects.get_for_object(obj))` much more
useful.

### DIFF
--- a/reversion/models.py
+++ b/reversion/models.py
@@ -62,6 +62,19 @@ class Revision(models.Model):
         help_text="A text comment on this revision.",
     )
 
+    def __repr__(self):
+        date_created = self.date_created
+        if date_created:
+            iso_date_created = date_created.replace(microsecond=0).isoformat()
+        else:
+            iso_date_created = None
+        return 'Revision(pk={!r}, comment={!r}, date_created={!r}, user={!r})'.format(
+            self.pk,
+            self.comment,
+            iso_date_created,
+            self.user,
+        )
+
     def revert(self, delete=False):
         # Group the models by the database of the serialized model.
         versions_by_db = defaultdict(list)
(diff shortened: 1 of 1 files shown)


blueyed:
I will add some tests in case you like it.

etianen:
Thanks for the pull request?

Is it normal to define custom __repr__ methods for models? I was under the impression that the Django built-in models typically don't. The only precedent I can find is with the admin LogEntry model.

blueyed:
It is not that common, but I've found it very useful in the past.
It gets used with locals in tracebacks and with the interactive interpreter.
Try the above example with and without this PR.

reversion/models.py
```diff
@@ -62,6 +62,19 @@ class Revision(models.Model):
...
+    def __repr__(self):
+        date_created = self.date_created
+        if date_created:
+            iso_date_created = date_created.replace(microsecond=0).isoformat()
+        else:
+            iso_date_created = None
+        return 'Revision(pk={!r}, comment={!r}, date_created={!r}, user={!r})'.format(
+            self.pk,
+            self.comment,
+            iso_date_created,
```
etianen: I think we can just use the built-in repr of date_created here.

reversion/models.py
```diff
@@ -299,6 +312,14 @@ def revert(self):
     def __str__(self):
         return self.object_repr
 
+    def __repr__(self):
+        return 'Version(pk={!r}, revision={!r}, object_id={!r}, content_type={!r})'.format(
+            self.pk,
+            getattr(self, 'revision', None),
+            self.object_id,
+            getattr(self, 'content_type', None),
```
etianen: Why are we using getattr?

reversion/models.py
blueyed: It is a bit long though - but has the benefit of being simpler.
Will change it.

reversion/models.py
blueyed: Because otherwise you might get an `RelatedObjectDoesNotExist` error if the ForeignKey is None.

PR #682, opened 2017-11-27
Travis: add codecov integration

blueyed:


etianen:
I've never heard of codecov before.

django-reversions does use `coverage` to generate code coverage. Does this have any advantage over that?

blueyed:
It is a coverage reporting platform, which provides reports like the one above (which is more useful once you have a base to compare against of course).
You can also browse the files and see the coverage at https://codecov.io/gh/etianen/django-reversion/tree/8412f1bfff056e627ee7044b6f50c68395299b26 already for example.

etianen:
I'm inclined to pass for now, as it's more complexity for little gain over the existing coverage solution.

blueyed:
Hmm.. but it does not replace the existing use of Coveragepy, but rather uses its output.
Therefore the patch to enable reporting to codecov (this PR) is just the reporting, and not e.g. installing/setting up coverage tracking (since that is already in place).
The only added complexity I see is that you might want to disable codecov comments (but would still benefit from its GitHub API status - e.g. it can report if a PR's diff is not covered enough with tests).

PR #708, opened 2018-03-20
set year properly

henning:
if 2.0.13 is released *after* 2.0.12 and assuming the day and month are right, it must have been released in 2018 not 2017...


henning:
didnt see before there is already a PR for this in https://github.com/etianen/django-reversion/pull/707

etianen:
Yup! #707 beat you to it. :)

PR #709, opened 2018-03-20
dynamic add delete model to Reversion

xiaxianzhi:
Hi, this PR is add deletion object into Version on signal pre_delete.  
 
Remember all deletion is important.
1. Most model objects are created in there own business views,  not django admin. Deleting objects in django admin may not be allowed, but hard to prevent. 
2. Command createinitialrevisions create reversion for all model records. So Version table becomes big and  it will cause slow sql and performance problem. Deletion is infrequent，import other independent objects into Version is not necessary.

### DIFF
--- a/reversion/revisions.py
+++ b/reversion/revisions.py
@@ -8,7 +8,7 @@ from django.core import serializers
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import models, transaction, router
 from django.db.models.query import QuerySet
-from django.db.models.signals import post_save, m2m_changed
+from django.db.models.signals import post_save, pre_delete, m2m_changed
 from django.utils.encoding import force_text
 from django.utils import timezone, six
 from reversion.compat import remote_field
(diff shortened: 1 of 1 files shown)


etianen:
If you create models in your own business views, use `RevisionMiddleware` to create the revisions, or even better, `reversion.create_revision()` decorator/context manager.

The trouble with this patch is that it will prevent fast-path deletion in Django (which is a lot faster!), and result in more revisions being saved in the usual case where models have a revision created on first save.

xiaxianzhi:
We use `django-reversion` to record **staff's modification** in django admin for audit as most developer does. Staff's modifcation and deletion is very less but unavoidable,  `django-reversion` can help us find out who made a mistake. 
 But model object changes in business view do not need to create reversion, especially in production environment.  We also have used logical deletion in business view when physical deletion in not allowed.
However, deletion is not recommended. The additional revisions of this patch(or older `django-reversion`) is much less than `createinitialrevisions `  or  `RevisionMiddleware `.  It is so expensive to support the fast-path deletion such an infrequent feature.

etianen:
You can implement this in your own project.

@receiver(pre_delete)
def post_delete_receiver(sender, instance, using, **kwargs):
    if reversion.is_registered(sender) and reversion.is_active() and not
reversion.is_manage_manually():
        reversion.add_to_revision(instance, model_db=using)

xiaxianzhi:
Thanks, we hava already done.

PR #714, opened 2018-04-23
Fix `ignore_duplicate_revisions` to `ignore_duplicates` in CHANGELOG.rst

myii:
With the former:

```python-traceback
TypeError: register() got an unexpected keyword argument 'ignore_duplicate_revisions'
```


myii:
Found a typo so added that commit to this PR as well.

etianen:
Good spot!

PR #722, opened 2018-05-19
Support for soft deletable models

### LINKED ISSUE #720: Support for soft deletable models
I recently run into trouble trying to use reversion within an app. The problem, I believe, was due to the fact that I was overriding the default manager of a model to filter some instances in a similar way as [SoftDeletableModel](https://django-model-utils.readthedocs.io/en/latest/models.html#softdeletablemodel) does. 
As far as I could track it, the error is located in the `_save_revision` method. Specifically the variable `model_db_existing_pks` is populated by checking the tracked objects' primary keys against those objects that exists in the DB. The problem is that this check is done by querying the DB using the `_default_manager` for each model. As you can guess, when the default manager filters out some instances, those instances' PKs will not survive the filter.
I wonder whether it would be more appropriate to use the `_base_manager` instead which is [not allowed to filter out any instances ](https://docs.djangoproject.com/en/2.0/topics/db/managers/#don-t-filter-away-any-results-in-this-type-of-manager-subclass).
This behaviour could at least be made explicit in the documentation. I believe this could save time to some other folks.
As you can see this is a question not an issue really.

ivissani:
Switched from _default_manager to _base_manager as _base_manger is, in theory, not allowed to filter model instances while _default_manager is.

This fixes a bug where updates to soft deleted model instances would not be saved into revision if _default_manager filter softdeleted instances out.

Closes #720


etianen:
Thanks!

I'm not sure the new model, migration and tests are necessary. If you could cull them, I'll merge.

ivissani:
I can cut them yet I think the test may be useful for 2 reasons:
a) It serves as documentation of why the change was introduced and
b) it serves as regression in case that any future change breaks the functionality.

If you are certain you want them cut, I'll do. Just let me know your final decision.

etianen:
Here's the thing - if every two line change in reversion generated 74
additional lines of test code, the situation would very rapidly become
unmaintainable!

I understand the argument, but test code has a tendency to grow
explosively, and each line of test code is another line in the codebase I
have to maintain.

ivissani:
Sorry for the late reply. I'm very busy these days. Will do what you asked and let you know.

etianen:
Thanks! It's merged.

Issue #725, opened 2018-06-02
Keep change_message as JSON


RamezIssac:
As suggested in #580 discussion 
We can keep change_message as JSON , and mimic admin.LogEntry change_message getter 

Goals are
1- Not violating the "no translated text" in database rule
2- More importantly, this change will yield correct change_message translation if language changed

PR #726, opened 2018-06-02
Leave change_message/comment as JSON

RamezIssac:
Addresses #725 

Accessing `reversion.comment` is now advised via `reversion.get_comment()` 
...should be documented and hinted in the release notes.

### DIFF
--- a/reversion/models.py
+++ b/reversion/models.py
@@ -7,6 +7,7 @@ try:
 except ImportError:  # Django < 1.9 pragma: no cover
     from django.contrib.contenttypes.generic import GenericForeignKey
 from django.conf import settings
+from django.contrib.admin.models import LogEntry
 from django.core import serializers
 from django.core.serializers.base import DeserializationError
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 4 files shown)


reversion/models.py
```diff
@@ -62,6 +64,47 @@ class Revision(models.Model):
...
+
+                elif 'deleted' in sub_message:
+                    sub_message['deleted']['name'] = ugettext(sub_message['deleted']['name'])
+                    messages.append(ugettext('Deleted {name} "{object}".').format(**sub_message['deleted']))
+
+            change_message = ' '.join(msg[0].upper() + msg[1:] for msg in messages)
+            return change_message or ugettext('No fields changed.')
+        else:
+            return self.comment
+
```
etianen: I'm not convinced by copying and pasting the code here.

How about you do this:

``` py
LogEntry(change_message=json.dumps(comment)).get_change_message())
```

This will re-use the existing code from Django quite cleanly.

RamezIssac: Yes,  absolutely right.
Updating

etianen:
Looks good, thanks!

I think this is going to require a major version bump to the library, as it's technically backwards-incompatible. Such is life! :)

RamezIssac:
Thanks to you for sharing and keeping up this wonderful app.

etianen:
I'll look into a 2.0 release in a week or so

etianen:
(I mean 3.0 release)

PR #728, opened 2018-06-13
When making a change during revert, create two revisions

matthijskooijman:
When you want to revert to an older revision, you get a regular change
form with the older values pre-filled. If you make changes to the form
before saving, the new version saved will be different from the older
version you are reverting to, though the log message still only says
"Reverted to previous version".

This commit changes the reverting code to create a separate revision for
the revert, with the "Reverted to previous version" message. The
subsequent save of the form happens in a second revision.

**This change is not complete**: Currently the second revision is always
saved, regardless of whether any changes were made. I am not sure how to
solve this yet.

I tried setting `ignore_duplicates` on the relevant model, which seems
to work. Of course this is something that should not be set for the
entire model, but just for this particular revision, but that is
probably feasible to add. However, `ignore_duplicates` seems to work on
individual versions, so might cause some versions to be left out of the
revision, even when the revision is not completely empty (more
generally, I believe `ignore_duplicates` might not always work as
intended when multiple models or follows are involved).

One other approach might be to to another check (like the one for
`ignore_duplicates`) just before saving a revision, but check *all*
versions in the revision and skip saving if *all* versions are
unchanged. This is probably a sane approach? The extra overhead of
comparing these versions is probably acceptable when it only happens on
a revert?

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -187,15 +187,17 @@ class VersionAdmin(admin.ModelAdmin):
         try:
             with transaction.atomic(using=version.db):
                 # Revert the revision.
-                version.revision.revert(delete=True)
+                with self.create_revision(request):
+                    version.revision.revert(delete=True)
+                    set_comment(_("Reverted to previous version, saved on %(datetime)s") % {
+                        "datetime": localize(template_localtime(version.revision.date_created)),
+                    })
                 # Run the normal changeform view.
                 with self.create_revision(request):
                     response = self.changeform_view(request, quote(version.object_id), request.path, extra_context)
                     # Decide on whether the keep the changes.
                     if request.method == "POST" and response.status_code == 302:
-                        set_comment(_("Reverted to previous version, saved on %(datetime)s") % {
-                            "datetime": localize(template_localtime(version.revision.date_created)),
-                        })
+                        pass
                     else:
                         response.template_name = template_name  # Set the template name to the correct template.
                         response.render()  # Eagerly render the response, so it's using the latest version.


etianen:
Are we sure that this is really a problem, in practice?

matthijskooijman:
I haven't used this much in actual practice yet, but I feel that the current message can be confusing and lacks explicitness. If you're looking at the history, you want to know what changed when, and should be able to trust the log that is shown.

matthijskooijman:
One alternative could be to make the form read-only somehow, so you can view a previous revision and revert to it, but not change it at the same time (this might be a useful feature in any case, to allow viewing history without reverting to it - even though that requires significant permission changes).

etianen:
Something to try: If we delete this line:

https://github.com/etianen/django-reversion/blob/master/reversion/admin.py#L196

Then I think that the normal django admin change message will be used,
which will simply generate a list of changed fields.

Sure, it'll no longer be displayed as a "revert" in the comment text, but
that's not really what's happening here anyway.

matthijskooijman:
>Then I think that the normal django admin change message will be used, which will simply generate a list of changed fields.

I believe it will only show the changed fields compared to the *reverted* version (since that is what the django admin classes believe to be the current version in the database). I guess this could be made to work if we only run this line for `POST` requests:

https://github.com/etianen/django-reversion/blob/master/reversion/admin.py#L190

Except that will prevent reverting fields that are not present in the change form, so that doesn't really help.

etianen:
I think maybe we go for a hybrid message. Something like "Reverted to
revisions saved on blah. Changed fields blah blah". Effectively,
concatenate the current revert message with the admin's generated change
message.

matthijskooijman:
That's actually a good idea. IIRC if no fields are changed (other than the revert), the admin generates a "No fields changed" messages, which is probably ok as well (though one could interpret the admin-generated message as the fields changed *during* the revert perhaps, rather than *after* the revert, but I'm not sure if there is an easy way to prevent that).

etianen:
Excellent! I'm happy to take a pull request for this idea.

PR #731, opened 2018-06-14
Add fields to context wrapper

massover:
This adds meta, user, and comment support to the `_ContextWrapper`.

I found it nice to build/test my own `create_revision`.

```python
def create_revision_for_obj(user, manage_manually=False, using=None, atomic=True, **kwargs):
    from reversion.models import Revision
    using = using or router.db_for_write(Revision)
    return _ContextWrapper(_create_revision_context, (manage_manually, using, atomic), ObjMeta, user, **kwargs)
```

I then could use it like so (as user and the meta obj were always required):

```python
meta_values = {'hello': 'world'}
with create_revision_for_obj(request.user, **meta_values):
  obj.save()
```

Compared to this:

```python
meta_values = {'hello': 'world'}
with create_revision():
  add_meta(ObjMeta, **meta_values)
  set_user(request.user)
  obj.save()
```

I'm not particularly attached to modifying the actual `create_revision` from the library. But I did add that in case that change to the api was liked. This is more about the changes to `_ContextWrapper` to build your own `create_revision` context managers. I am glad to remove it.

### DIFF
--- a/reversion/revisions.py
+++ b/reversion/revisions.py
@@ -297,30 +297,47 @@ def _create_revision_context(manage_manually, using, atomic):
         _pop_frame()
 
 
-def create_revision(manage_manually=False, using=None, atomic=True):
+def create_revision(manage_manually=False, using=None, atomic=True, meta=None, user=None, comment=None, **kwargs):
     from reversion.models import Revision
     using = using or router.db_for_write(Revision)
-    return _ContextWrapper(_create_revision_context, (manage_manually, using, atomic))
+    return _ContextWrapper(_create_revision_context, (manage_manually, using, atomic), meta, user, comment, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


massover:
@etianen that was a great suggestion!

I moved the code to `create_revision_context`. I actually removed the changes from create_revision as to not touch that since I don't need it. I also added an arguably useless test in another commit so it can be removed easily if you prefer to not keep it. Also not sure if there is any documentation to update for this 🤷‍♂️ .

etianen:
Thanks!

Any chance of a pull request to update the docs: https://github.com/etianen/django-reversion/tree/master/docs ?

etianen:
Unfortunately, I've had to revert this pull request.
Extra fields should be added to create_revision(), not _create_revision_context(), which is a private API.

PR #733, opened 2018-06-20
Allow overriding of `_request_creates_revision`.

kklingenberg:
Through the custom methods `RevisionMixin.revision_request_creates_revision` and `RevisionMiddleware.request_creates_revision`, as well as an additional optional parameter for the view decorator `create_revision`, the user can provide a predicate that may extend or override the default "reject if GET, HEAD or OPTIONS" criterion.

I'm aware that this behaviour can be achieved by not using the mixin / middleware / view decorator and manually annotating blocks of code. However, I believe this would simplify things by centralizing any additional, all-covering criteria.

### DIFF
--- a/docs/middleware.rst
+++ b/docs/middleware.rst
@@ -9,14 +9,14 @@ Shortcuts when using django-reversion in views.
 reversion.middleware.RevisionMiddleware
 ---------------------------------------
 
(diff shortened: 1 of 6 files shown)


etianen:
I like it!

Can you update the documentation with this new hook? I'll then merge.

etianen:
Sweeeeeet!

Issue #735, opened 2018-06-22
How to store the timestamp and author of a delete operation?


kklingenberg:
I'm requesting advise :grimacing:. I need to be able to tell who deleted an object, and when it happened. I'm considering three options:

1. Fork and add the `post_delete` signal that would create a revision, where the corresponding version would be tagged as a delete. An attempt at this can be seen [here](https://github.com/kklingenberg/django-reversion/commit/88a35040a927922d050ce9d228d3700d2caa4f4d). However, I believe this goes against the scope of django-reversion, since the `get_deleted` queryset method already deals with half of what I need, and this would overlap with it somewhat.
2. Inherit from `reversion.models.Version` and add a foreign key to the user (the author of the deletion) and an extra timestamp, so that a previous version can be tagged as 'the last one before a delete' and can be annotated with what I need. I'd also need to handle the `post_delete` for each registered model. I dislike this because the table inheritance will create another version table.
3. Fork and allow users to provide their own version model, by making `Version` abstract and providing a default concrete model. This would require app configuration though, and I think that's also a major change from the current status.

I'd appreciate any guidance on this, and am willing to submit a PR for whichever approach you're willing to include in the library, if any.

etianen:
There's no need to add a foreign key from Version to User. Such a foreign
key already exists on Revision, so you should use that.

A good solution is to save a revision in the pre_delete signal of a model,
and tag that revision with the user. This means that you can tell who
deleted the object by looking at the user associated with the last revision
of the object.

kklingenberg:
Aha! I see :smile: 

Thank you very much! I was blocked on tagging the version as deleted, but `VersionQuerySet.get_deleted()` does that

depaolim:
I connected a pre_delete signal for the model
In the connected function I issued an explicit reversion.add_to_revision(instance)
But I can't get any stored version/revision
The reason is that reversion._save_revision "Only save versions that exist in the database." (as the comment states)
Am I missing something?
Thank you

kklingenberg:
I checked again and found your problem (and a bug in my setup). What I was doing was:

```python
def pre_delete_receiver(sender, instance, **kwargs):
    with reversion.create_revision():
        reversion.add_to_revision(instance)
```

[...]

Issue #164, opened 2012-06-19
Remove the added / updated / deleted flag on version models


etianen:
This feature was added to django-reversion in 1.4, in order to allow statistics to be gathered about additions / updates / deletes for apps that required them.

Well, turns out it was a pretty bad idea.

Prior to this, reversion had a simple, git-like model whereby Revisions represented complete snapshots of a related set of models at a point in time. The idea of them corresponding to additions / updates / deletions wasn't really relevant. The first revision mentioning a model represented it's "addition", and the last revision mentioning a model that no longer had a representation in the database represented it's "deletion". Recovering a deleted model was a simply case of instantiating it from it's last mentioning Revision.

Importantly, **when a model was deleted, no Revision would be created**.

By adding in the delete flag, it became a necessity to save a new revision when a model was deleted. However, when considering foreign relations, this opened up a whole can of worms.

The most obvious of these is bloating the revision/version database, but that's not the biggie.

Consider the following case:

``` python
class Person(models.Model):
    name = models.CharField(max_length=20)

class Pet(models.Model):
    owner = models.ForeignKey(Person)
    name = models.CharField(max_length=20)
```

revision 1:
- create person1
- create pet1 (owner person1)

revision 2:
- delete pet1

Asking reversion to list all revisions for person1 will return these two revisions. Reverting to revision 2 will be a no-op. Reverting to revision 1 will recreate pet1.

Asking reversion to list all revisions for pet1 will return the save two revisions. Reverting the whole revision will be a no-op, since you're asking reversion to return the models to a state where pet1 was deleted. However, if you just revert the Version model, you'll get a pet back.

In other words:

``` python
pet_version.revert()  # This will revert pet 1.
pet_version.revision.revert()  # This will not revert pet 1.
```

Complicated enough? Now it's time to consider the get_deleted method. In the current implementation, if you were to call get_deleted() on the Pet model, the version it returned would correspond to revision1, which _isn't the latest revision in the database_. This is because get_deleted is trying to return the revision that would allow the given model's revision to be recreated. If you were trying to access meta information stored at the time the model was deleted, then you have to use reversion.get_for_object_reference() instead.

**And all this complexity just to make gathering statistics on number of deletes in the database easier?!**

The madness has to end. When gathering statistics has made the whole of reversion much more complicated to understand for it's core functionality, something has gone wrong.

PR #738, opened 2018-06-30
Optionally track deleted objects into a revision.

kklingenberg:
Provide the option to register a model for 'deletion-tracking' (disabled by default, for backwards-compatibility), in which the `pre_delete` signal is listened to, and a consequent version is created. The version won't be saved if there's no previous revision history for said object.

This is an attempt at dealing with #735

I'll gladly write more tests and document the corresponding API changes if there's any merit to this PR.


etianen:
Thanks for the pull request!

Unfortunately, tracking deletions is something that doesn't really work with django-reversion. There's a discussion about it here: https://github.com/etianen/django-reversion/issues/164

PR #742, opened 2018-07-20
Add extra fields to create_revision interface

massover:
This is a continuation of https://github.com/etianen/django-reversion/pull/731

> Unfortunately, I've had to revert this pull request.
> Extra fields should be added to create_revision(), not _create_revision_context(), which is a private  API.

If I understood this correctly, I added the the feature to the `create_revision` as that is the public api. I did also have to modify `_create_revision_context()` in order to allow the `create_revision` update to work, as per your original suggestion:

> Nice idea!
> 
> I think that this should be implemented inside the _create_revision_context function, since then the calls to set_meta, set_comment etc only need to be implemeted in one place.

Also added documentation.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -146,6 +146,22 @@ Reverting a :ref:`Revision` will restore any serialized model instances that hav
     assert obj.name == "version 2"
 
 
+Custom create revision context managers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Create a `create_revision` context manager that is specific to your application and models.
+
+.. code:: python
+
+    def create_revision_for_your_obj(user, status):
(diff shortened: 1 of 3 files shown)


etianen:
Thanks!
I'll review this when I get back from holiday, in just over a week. :)

etianen:
Thanks!

Looking at your use-case (custom create_revision context managers), it looks like you can do this already without changing django-reversion at all!

``` pyy
from contextlib import contextmanager
from reversion import create_revision, set_user, add_meta

@contextmanager
def create_revision_for_your_obj(user, status):
    with create_revision():
        set_user(user)
        add_meta(YourObjMeta, status=status)
        yield

with create_revision_for_your_obj(request.user, 'draft'):
        obj.save()

``

massover:
Perfect. I didn’t think to simply nest the context managers!

PR #756, opened 2018-09-25
Fixed TypeError exception when accessing m2m_data attributes

denisroldan:
Even if you have a M2M relation to `django.contrib.auth.models.User` or you call `django.contrib.auth.get_user_model()` you can’t access `Version.field_dict` attributes.

In my case I had a Model like this:
```
@reversion.register()
class MyModel(models.Model):
    users = models.ManyToMany(get_user_model(), related_name='my_models', blank=True)
    
    # More fields
    (...)
```

When I was working on a low-level codebase around Version objects, I found that calling `version_obj.field_dict[field_name]` doesn't works if a M2M was present pointing to User. Any other M2M relation works smoothly. 

https://github.com/etianen/django-reversion/blob/3d05eed4c6e242bf7f0e8bd4726918bb3a771764/reversion/models.py#L269-L274

The problem is that `m2m_data` is not always available and it raises a `TypeError: argument of type 'NoneType' is not iterable`.

This PR fixes that problem :)

### DIFF
--- a/reversion/models.py
+++ b/reversion/models.py
@@ -270,7 +270,7 @@ class Version(models.Model):
             field = model._meta.get_field(field_name)
             if isinstance(field, models.ManyToManyField):
                 # M2M fields with a custom through are not stored in m2m_data, but as a separate model.
-                if field.attname in object_version.m2m_data:
+                if object_version.m2m_data and field.attname in object_version.m2m_data:
                     field_dict[field.attname] = object_version.m2m_data[field.attname]
             else:
                 field_dict[field.attname] = getattr(obj, field.attname)


etianen:
Can I see the full stack trace of the error, before I merge this? Just to verify the cause of the original problem.

denisroldan:
Thanks for the quick response! For NDA reasons, I just changed some folder names :) 

Here's the full traceback:
```
Traceback (most recent call last):
  File "/opt/python/3.6.3/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/python/3.6.3/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/home/travis/build/xxx/xxx/xxx/tests.py", line 92, in test_custom_diff_compare
    diff = mr.get_diff()
  File "/home/travis/build/xxx/xxx/xxx/models/versioned_object.py", line 57, in get_diff
    diffs = dict(generate_diffs(self.prev_version, self.next_version, field, 'semantic'))
  File "/home/travis/build/xxx/xxx/xxx/utils/reversion_diff.py", line 10, in generate_diffs
    old_text = old_version.field_dict[field_name] or ""
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/utils/functional.py", line 35, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/reversion/models.py", line 295, in field_dict
    field_dict = self._local_field_dict
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/django/utils/functional.py", line 35, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/reversion/models.py", line 281, in _local_field_dict
    if field.attname in object_version.m2m_data:
TypeError: argument of type 'NoneType' is not iterable```

Custom code is about creating a diff, inspired on existing diff work and `django-reversion`. None of them was what we need so I was working on a custom code for that and the problem appeared.

etianen:
Very odd, and I suspect a bug in Django.

Meanwhile, your patch does the job. Thanks for contributing!

denisroldan:
You're welcome! Yep, it was really strange. Thanks for merging :)

Issue #699, opened 2018-01-27
Squashed migrations


sximba:
Will the squashed migrations (i.e 0001, 0002, 0003, 0004) ever be deleted as advised in the django docs?

The squashed migration has existed for 2 years now. I think it is safe to assume everyone who uses the project has applied it.

Their presence in the code base makes it impossible to unapply migrations and leave the django migration tree clean.

etianen:
Good point. This has been done in master.

BallisticPain:
Yet we're still having issues where people are NOT ready for them to be gone. I've copied them into a separate directory myself... :( Having to handle it manually... Oh well I'll be good after this... :-D

luzfcb:
I was bitten by it. Will be included in the changelog in which version was removed.

etianen:
Okay... I won't be removing squashed migrations any more. I'll even take a pull request to put them back in.

merwok:
Can confirm than doing a big update on a existing live project after two years does happen.  Current migration is 0002_auto_20141216_1509 :)

etianen:
**I'll even take a pull request to put them back in.**

libcthorne:
I hit the same problem. PR here, as suggested: #757

etianen:
Pull request merged. If people could confirm that it's working for them on master, I can make this into a release. Thanks @libcthorne

libcthorne:
I can confirm `master` works for me:
```
Running migrations:
  ...
  Applying reversion.0003_auto_20160601_1600... OK
  Applying reversion.0004_auto_20160611_1202... OK
```
A release would be appreciated.

etianen:
Released 3.0.1

etianen:
This has broken the upgrade from v3.0.0. It looks like it's impossible to add squashed revisions back in without triggering some broken behavior in the django migrations framework. See #763

PR #757, opened 2018-10-22
Re-add old squashed migrations for smooth upgrade from older versions…

libcthorne:
We hit a missing migration issue while upgrading from version 1.10.2 to 3.0.0 due to the squashed migrations being removed. This PR re-adds them as suggested by [this comment](https://github.com/etianen/django-reversion/issues/699#issuecomment-428851466).


etianen:
You da MVP. :)

PR #766, opened 2018-11-02
Added support for clickable urls for admin user who made edits.

prakhar-goel:
This is how it would look:
screen shot 2018-11-02 at 11 18 59 pm

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 import json
 from contextlib import contextmanager
+from django.conf import settings
 from django.db import models, transaction, connection
 from django.conf.urls import url
 from django.contrib import admin, messages
@@ -8,6 +9,7 @@ from django.contrib.admin import options
 from django.contrib.admin.utils import unquote, quote
 from django.contrib.contenttypes.admin import GenericInlineModelAdmin
 from django.contrib.contenttypes.fields import GenericRelation
+from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import PermissionDenied, ImproperlyConfigured
 from django.shortcuts import get_object_or_404, render, redirect
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


etianen:
I think that this is useful, but harder to do in practice than it may first
appear.

My concerns:
- It'll break if there is no user registered in the django admin.
- It'll break if the user doesn't have permission to view the user model.

The default object history view in the django admin doesn't include user
URLs, probably for this reason.

prakhar-goel:
Hi Dave,


> - It'll break if there is no user registered in the django admin. 

I have added a check for this in template. Following suggestions from this question https://stackoverflow.com/questions/694477/getting-django-admin-url-for-an-object
```
...
{% url admin_user_view action.revision.user.id as admin_user_url %}
{% if admin_user_url %}
...
```
Please double check if this will remedy the problem.

> - It'll break if the user doesn't have permission to view the user model.

If a user has permission to view the log history (which usually contain more sensitive data), there is a high chance that user has permission to view user model as well. In case he doesn't, they would see a 403 page. This is more of sanity than a break in flow. Even in AWS console, when I login as IAM user, I regularly see "permission denied" although I can see links to forbidden resources.

RamezIssac:
> If a user has permission to view the log history (which usually contain more sensitive data), there is a high chance that user has permission to view user model as well.

I disagree here .. having permission to view a model (ex: Product) history , does not mean i can see my "manager" profile .. 
Also this is not view page you're linking to.. it's a the change page.
Finally: If it'll throw a 403 then maybe it's not OK to have it as a link in the first place.

My suggestion:  give just enough support for this option and leave it to the end developer to actually implement it.

etianen:
I'm inclined to agree. On reflection, the object-history page in django-reversion should be as close to the equivalent raw django page as possible. It was done that way for a reason.

User's can implement this themselves very easily.

PR #768, opened 2018-11-14
allow users with view permissions to view history

ofw:


ofw:
Without this, users that have only view permissions get 403 when pressing "History" button on a model's page. But they can still view a revision if they get a direct url to it.

ofw:
this won't work for django < 2.1 though

etianen:
A very nice change, and good work on backwards compatibility!

PR #774, opened 2019-01-23
Update safe_subquery method to use exists as it is much much faster

MRigal:
The test suite itself runs in 4 instead of 7 seconds on my local machine.

The NOT IN that was previously used grows exponentially with the data size, whereas the NOT EXISTS is almost flat. The code is already more than 1000x faster when running a createinitialrevisions with 50000 objects in PostgreSQL.

Only constraint, it is Django 1.11+ only, but so is already at least the test suite, isn't it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,8 @@ services:
   - postgresql
   - mysql
 install:
-  - pip install flake8 coverage mock sphinx django$DJANGO psycopg2 mysqlclient -e .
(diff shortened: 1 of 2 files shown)


MRigal:
I fear the failing tests are not related with my changes. Also master was failing there. Thus I changed the travis config accordingly

etianen:
This is fantastic. Thanks!

PR #776, opened 2019-01-29
fix version view for model inherited from other model

ofw:
Now when one has:

```
model A(models.Model):
   ....

model B(A):
   ....
````

And try to view version for model B, exception occurs:

"AttributeError: 'HttpResponseRedirect' object has no attribute 'render'" `line 196, in _reversion_revisionform_view`

That's because `current_revision` contains both A and B instances while `old_revision` contains only B instance. Then instance A is deleted but because of model inheritance and one to one field, instance B is also deleted from db.

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -170,7 +170,7 @@ class VersionAdmin(admin.ModelAdmin):
         try:
             with transaction.atomic(using=version.db):
                 # Revert the revision.
-                version.revision.revert(delete=True)
+                version.revision.revert(delete=False)
                 # Run the normal changeform view.
                 with self.create_revision(request):
                     response = self.changeform_view(request, quote(version.object_id), request.path, extra_context)

### CODE REVIEW

**Comment by ofw** [comment]
oh, some tests are failing now

**Comment by ofw** [comment]
now it works, but please tell me if something may go wrong. I disabled deleting models from current revision. Later code under `if delete:` should be deleted because it is not used anymore

**Comment by ofw** [comment]
Any comments? Otherwise I'm going to close this MR.

**Comment by etianen** [comment]
The delete functionality in the admin revert is pretty important behaviour, IMO. I don't want to simply disable it.

I'd take a merge request that fixes the original bug.

**Comment by somada141** [comment]
@etianen @ofw I'm hitting this error as well, in particular on the combination of `django-reversion` with [`django-polymorphic`](https://github.com/django-polymorphic/django-polymorphic) though I dunno if the latter influences the behaviour. 

Effectively, when I attempt to view a deleted record in order to recover it I'll hit a 500 with the same error as above and this traceback:

```
AttributeError: 'HttpResponseRedirect' object has no attribute 'render'
  File "django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "django/contrib/admin/sites.py", line 231, in inner
    return view(request, *args, **kwargs)
  File "reversion/admin.py", line 201, in recover_view
    return self._reversion_revisionform_view(
  File "reversion/admin.py", line 178, in _reversion_revisionform_view
    response.render()  # Eagerly render the response, so it's using the latest version.
```

at which point if i hit back on the browser the list-view of the deleted items will show up with a warning  at the top of the page reading `<verbose-name-of-model> with ID “<pk>” doesn’t exist. Perhaps it was deleted?`.

Apologies for commenting on a closed issue but I was curious as to whether either of you have managed to resolve this issue in any way eg properly redirecting to the list-view.

etianen:
I think using model inheritance with the admin integration is going to land you in problems, as it effectively means you've got two ways of saving a given model. I'm not sure of any easy way to resolve it.

somada141:
Yeah I thought the combination of polymorphism with reversion might be a bit of a hornet's nest (honestly I'm surprised it has worked as well as it has thus far). I've been investigating the issue which has not been super reproducible as it seems to have something to do with whether the view called when viewing a deleted item is that of the child or the parent object.

I will look further into it and will update here should I manage to resolve it somehow

PR #782, opened 2019-03-20
Refactor Readme to use Pypi badges

rhenter:
Add Pypi badges and put in the head. 

If you want to see how is rendered after the change follows the link below:

https://github.com/rhenter/django-reversion


etianen:
That looks great! Thanks.

Issue #557, opened 2016-07-18
Standardize on JSON serialization format


etianen:
- [ ] Use JSONField for serialization of data in Postgres.
- [ ] Remove `serialization_format` column.

Pros:
- Faster deserialization in postgres.
- Allows querying the serialized data in postgres.

Cons:
- High storage overhead (binary JSON takes up more disk space than text!)
- Some people like non-JSON serialization format, for some reason.
- A fairly expensive database migration for people using JSON serialization.
- An incredibly expensive database migration for people using non-JSON serialization.
- This would require a v3.0.0 release of django-reversion.

danielsamuels:
It's worth noting that 1.8 LTS is going to be supported for at least another 2 years and `JSONField` wasn't added until 1.9, so this might have to wait until the 1.8 LTS ends.

etianen:
A good point. This isn't an urgent feature, by any means.

eronlloyd:
Is it possible to make this an option now, while not yet the default? How is JSON currently stored?

etianen:
JSON is currently stored as a TextField, which makes it pretty impossible to query effectively.

Then again, in postgres, you could query it by using custom SQL and casting the field to json. That would let you use all the JSON operators. It would be slow, but querying over JSON is going to be slow anyway.

etianen:
Hmm, in fact, the more I think about it, the less this seems like a good feature.

- You can already do JSON lookups in postgres using custom SQL. All this feature would add is the ability to use Django's JSON database operators.
- Binary JSON takes up more disk space, and django-reversion is already a disk hog.
- Querying binary JSON is still very slow, unless you add a GIST index. Adding a GIST index will make it take up even more space on disk!

sliverc:
I would also love to see data be stored as json field mainly to be able to query json data easily.

As django-reversion should work on other databases then postgresl it is going to be tricky to support `JSONField` for Postgres but fall back to `TextField` on other databases in the current state of Django.

This [discussion](https://groups.google.com/forum/#!topic/django-developers/zfred27yVPg) might be of interest whereas a generic json field in Django would make life easier and more consistent.

MRigal:
@etianen now that Django 3.2 brings json storage support to all databases, the time might has come to allow this to happen! :-)

etianen:
I'm in two minds. A lot of people have huge reversion databases, and the migration to support this will take hours and hours to run. That seems risky IMO.

[...]

PR #790, opened 2019-05-07
Make serialized column into JSONField for Postgres

tkshnwesper:
Hello,

Just proposing a probable solution for #557

Cheers!

### DIFF
--- a/reversion/models.py
+++ b/reversion/models.py
@@ -10,7 +10,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core import serializers
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.serializers.base import DeserializationError
-from django.db import IntegrityError, connections, models, router, transaction
+from django.db import IntegrityError, connection, connections, models, router, transaction
 from django.db.models.deletion import Collector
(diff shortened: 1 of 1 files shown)


etianen:
This won't work, sadly. Depending on the database where this is run, it'll generate different migrations files.

etianen:
I think the correct approach is *probably* a custom model field that is stored as bjson in postgres, and text everywhere else. I'm not sure that this complexity is worth slightly easier querying in postgres. It should be possible to run any JSON query in postgres using custom SQL, and casting the field into json. The performance won't be great without an index, but it would avoid a lot of complexity in django-reversion itself.

tkshnwesper:
Ah yes, I didn't take into consideration the different migration scripts that would get generated.

> It should be possible to run any JSON query in postgres using custom SQL, and casting the field into json.

I see what you mean, but the selectors that Django provides for `JSONField`s might be better to use when it comes to speed (since internally the field is a `jsonb` and hence indexed) and for better readability of code.

But I'm not very sure if a text search on a `TextField` would be faster than querying for a key-value on a `JSONField` column, but I have a hunch that the text search might be faster.

Thanks

PR #801, opened 2019-09-10
Drop Python 2 support

claudep:


claudep:
This will be mandatory to support Django 3.0 (unless we add more compatiblity shims).

etianen:
Very nice. I suspect there'll be a few more places where the code can be cleaned up without the requirement to support Python 2, Django < 3.0, but this is a great start.

I think that following Django's advice to drop Django <3.0 support and Python 2 probably merits a breaking release (v3.0?), so if there's anything else you can see that will cut out compatibility shims, I'd love a PR.

claudep:
For now I didn't see yet a compelling reason to drop support for older Django version, but it might occur when adding support for Django 3.0.

PR #804, opened 2019-09-11
Django ticket 27218 was implemented in Django 1.11

claudep:


etianen:
Nice. All your recent PRs are merged.

I can't see any reason to release yet, since they're largely code simplifications in preparation for Django 3.0. When Django 3.0 lands, the recommendation is to drop support for Django < 2.2, which I guess we should follow, and make a new release.

https://docs.djangoproject.com/en/dev/releases/3.0/

claudep:
I must admit I don't fully agree with that recommendation. For some packages where you have to add semi complicated code to support older versions, yes sure. But typically in this case, we can easily support 3.0 and even 1.11 without any issue, so I wouldn't press for dropping those versions too soon. But you are the maintainer :-)

PR #806, opened 2019-10-30
Allow customize comment

ehooo:
Add new function to allow customize comments.

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -61,7 +61,7 @@ class VersionAdmin(admin.ModelAdmin):
         return queryset
 
     # Messages.
-    
+
     def log_entry_to_comment(self, entry):
         return entry.get_change_message()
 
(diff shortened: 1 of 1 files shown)


etianen:
Hi! Thanks for the MR.

I think there's already a hook on the admin class for precisely this purpose:

`construct_change_message`

Alternatively, the user can just override `log_addition` and `log_change` on their admin class and substitute in a different message.

ehooo:
Hi @etianen i already see it, but this is for customize the content on the comment it self.
The text that you could see on the web page.

I want store the json format on the LogEntry and see text friendly on the comment in the reversion model.
In that way i could have alot information on the json inside of the LogEntry model and summary on the reversion model.
I want get the LogEntry to analyze everything with details, in other way i cannot have different information on each side.

etianen:
That sounds like a pretty unusual use case. People using django-reversion
typically disregard LogEntry.

Rather than add overhead to the library, I think this is best solved in
your own code.

PR #812, opened 2019-12-12
Emphasise that clicking the link doesn't revert

Flimm:
A small change, it would be quicker to view the diff than to explain it.

As a new user, when I first saw the sentence, I thought that clicking the link immediately reverted to that version, when in actuality, it displays the version first.

### DIFF
--- a/reversion/templates/reversion/object_history.html
+++ b/reversion/templates/reversion/object_history.html
@@ -5,7 +5,7 @@
 {% block content %}
     <div id="content-main">
 
-        <p>{% blocktrans %}Choose a date from the list below to revert to a previous version of this object.{% endblocktrans %}</p>
+        <p>{% blocktrans %}Choose a date from the list below to view a previous version of this object and to revert to it.{% endblocktrans %}</p>
 
         <div class="module">
             {% if action_list %}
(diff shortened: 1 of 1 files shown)


reversion/templates/reversion/object_history.html
```diff
@@ -5,7 +5,7 @@
 {% block content %}
     <div id="content-main">
 
-        <p>{% blocktrans %}Choose a date from the list below to revert to a previous version of this object.{% endblocktrans %}</p>
+        <p>{% blocktrans %}Choose a date from the list below to view a previous version of this object and to revert to it.{% endblocktrans %}</p>
```
etianen: I see what you're doing, but I'd say this description still implies that clicking the link will revert immediately!

How about "Choose a date from the list below to view a previous version of this object, and optionally revert to it."

etianen: Of course, this will break all translations, so I'm not sure it's worth the clarification, since my translations are unlikely to get updated for a while.

Flimm: I updated the text. It's up to you to decide whether it's worth touching the translations.

etianen:
Thanks!

I'm really torn here. I'd assume that the majority of users are English-speaking, so this is to the benefit of the majority of users. I might sit on it for a bit and have a think.

PR #814, opened 2020-01-07
Replace force_text with force_str

Tenzer:
The former is deprecated in Django 3.0 and has been an alias for force_str
since Django 2.0:
https://docs.djangoproject.com/en/3.0/releases/3.0/#django-utils-encoding-force-text-and-smart-text

Use of force_text raises deprecation warnings on Django 3.0.


etianen:
Thanks!

PR #816, opened 2020-01-13
Properly support revisions in on_commit handlers

sigvef:
Before this commit, making a second revision after a first atomic
revision transaction has successfully committed does not work properly.
Now, making a new atomic revision inside an on_commit handler works as
expected.

Implementation-wise it looks like it was already supposed to work, but
the internal local stack bookkeeping pops the current frame off the
stack just too late when exiting the contexts. This tricked the
bookkeeping into skipping the second revision creation.

### DIFF
--- a/reversion/revisions.py
+++ b/reversion/revisions.py
@@ -277,6 +277,7 @@ def _dummy_context():
 @contextmanager
 def _create_revision_context(manage_manually, using, atomic):
     _push_frame(manage_manually, using)
+    expected_stack_length_after_pop = None
     try:
         context = transaction.atomic(using=using) if atomic else _dummy_context()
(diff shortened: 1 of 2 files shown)


etianen:
Thanks for this! I've kept your test, but written an alternative implementation in #818 that keeps things simpler, and seems to acheive the same thing my simply reversing the order of the nesting of atomic blocks and revision stack frames. Your new test still passes.

Does #818 make sense to you?

sigvef:
Yes, that seems like a good way of fixing it. Thanks!

etianen:
Cool. I'll merge #818, and close this. Thanks for the MR and the review :D

PR #817, opened 2020-01-14
Adjust setup

michael-k:
- Enforce Python 3.6+
- Make wheels non-universal (Python 2 support is missing for universal)
- Add a long description for https://pypi.org/project/django-reversion/


etianen:
That's cool, thanks!

Keeping up with packaging specifiers is tiresome, so I appreciate it! :)

PR #832, opened 2020-04-20
Use ISO 8601 date format to disambiguate dates

annainfo:
3.0.2 - 05/11/2018 can refer to May 11 and 5 Nov

https://www.iso.org/iso-8601-date-and-time-format.html

https://www.iso.org/standard/15903.html - ISO 8601 exists since 1988 

https://www.python.org/ uses this format in Latest News and Upcoming Events

https://xkcd.com/1179/


etianen:
I mean, you're totally right. Thanks!

Issue #822, opened 2020-02-15
Removing squashed migrations prevents migration


jribbens:
Because the migrations have been squashed, migration is impossible as it returns the following error: `django.db.utils.ProgrammingError: relation "reversion_revision" already exists`... you have to `pip install django-reversion==3.0.1 && ./manage.py migrate reversion && pip install -U django-reversion`

If there were only 4 of them and somehow having both the originals and the squashed one causes problems, why not just remove the squashed one and keep the originals?

etianen:
Once a set of migrations have been squashed, they cannot be un-squashed without breaking existing projects. It's too late.

It should work to upgrade to 3.0.0, run `migrate`, then upgrade to `3.0.5`, then migrate again.

m4rk3r:
Perhaps this could be better called out in the changelog for those of us upgrading from old versions? @jribbens solution (uninstall, install 3.0.1, migrate, upgrade) worked

etianen:
Happy to take a MR to add a note to the changelog.

PR #833, opened 2020-04-29
Update Changelog to add upgrade note

m4rk3r:
When upgrading from old versions it's necessary to incrementally upgrade at version `3.0.1`, run migrations, and then proceed.

[See here for more info](https://github.com/etianen/django-reversion/issues/822#issuecomment-621361180)


etianen:
Clear and concise. Love it!

Issue #773, opened 2019-01-20
Tests runnable without external database?


risicle:
It would be particularly neat if the test suite's external database tests were optional. This would allow the django-reversion Nix package (https://github.com/NixOS/nixpkgs/blob/ca7ef865c45726dc80ba6e3d42e37ccfa6c0ae73/pkgs/development/python-modules/django_reversion/default.nix) to enable the tests, which would give package maintainers bumping the version a little more confidence they're not building broken packages. (They'd also get re-run whenever any of its dependencies get bumped to check *those* haven't broken it)

It does appear that a good few of the tests simply work against the default sqlite database and it would probably just be a matter of scattering some `@skipIf` decorators or `skipTest()` calls throughout the others. For bonus points, make `settings.py` simply omit the missing entries of `DATABASES` if the relevant db credentials e.g. `DJANGO_DATABASE_NAME_POSTGRES` aren't found in the env.

risicle:
I'm confused... unless I'm reading the code wrong, ~70% of your tests don't touch the postgres or mysql databases and so "aren't replicating a production scenario" - surely us checking 70% of tests pass would be better than the current situation where the package could get *totally* broken and nobody would be any the wiser?

etianen:
Looking closer at the Nix Django package... that doesn't seem to be running
tests either...

risicle:
No, that's actually been a project of mine that I haven't got a lot of traction on - it has thousands of tests. In the mean time it's a lot better to try and get *something* for as many of the smaller projects as possible (in a way, ensuring tests pass for packages that *depend on* django gives an amount of assurance about django itself).

risicle:
Anyway, I was only suggesting a ~20 line change that would have been kinda helpful to packagers, let's not get too caught up in it.

PR #835, opened 2020-05-12
Docker

Martini991:
I added some configurations to simplify testing with docker.
I also added some docs in the README.rst.

Lorenzo


Martini991:
I looked at the open issues: #773 can be considered closed?

etianen:
I appreciate the work you've put into this, and it's a nice idea. Unfortunately, it's quite a weight of files to add to the project for what's effectively an alternative testing workflow, and since it's not a workflow I use myself, there'll be a tendancy for it to rot and break without anyone noticing.

Any open-source project can be tested using Docker. But should every open source project include a load of stuff to make doing so easier? To my mind, no.

Thanks for the MR. I can't accept it at this time, but it's a nice piece of work, and the documentation additions were well written. :smiling_face_with_three_hearts:

Martini991:
Hey @etianen! Thanks for your response! 😄

If you don't want to merge my pull request its fine, but I think its a pity because without those scripts making a contribution is way harder.

Take myself for example, I wanted a demo of the functionalities STAT and I had to install **two DB** on my machine to do so, and without my work someone else have to do the same thing **again**! Let's not repeat ourselves!

An open source project lives with his community, and if someone gives up on a contribution because its too much pain to set-up, then it's a loss for everyone. 😣

Also I have two questions about your comment:

**1.**
> it's quite a weight of files to add to the project

What do you mean exactly? Not considering the docs and the 4 lines edits in "test_project/settings.py" the wc is pretty low:
```bash
  $ wc -l Dockerfile docker-compose.env docker-compose.yml docker-entrypoint.sh docker-run.sh
        35 Dockerfile
        22 docker-compose.env
        41 docker-compose.yml
        20 docker-entrypoint.sh
        13 docker-run.sh
       131 total
```
Maybe the fact that everything is in the root folder bothers you? I can move everything in a "docker" subdir if you feel that's cleaner, otherwise mind to explain the issue in more detail?

**2.**
> it's not a workflow I use myself, there'll be a tendancy for it to rot and break without anyone noticing.

Why don't you start using it and dismiss the old method that relies on the underling OS with its DBs? What's stopping you? Is there some hidden advantage I don't see?

If you don't want to spend a lot of time writing a response it's fine (and understandable) so I also throw in the option to chat via Skype (or whatever you prefer), just comment here and I will send to your e-mail (dave@etianen.com) my contact.

Anyway, thank you for your time and your compliments about my MR! 🤩

Lorenzo

etianen:
I generally assume that people would demo django-reversion on an existing project. Adding it to an existing project is just a few lines of code, and then you get to see exactly how it would work with what you have. Even better, if you like it, you've just done the work to integrate it! :p

In terms of testing and contributing, your point is well-made. However, it's a choice of installing two databases **or** docker. On Unix, that's the same difficulty, as it's a single `apt-get` command to do both. On OSX, if you have homebrew, it's the same. On Windows, I have no idea! But not everyone has Docker, and not everyone wants to install it. It probably does lower the barrier to entry for Docker users, but for everyone else it's more docs, more files to understand.

Don't get me wrong, I'm not a Docker-hater. I use it a tremendous amount in my day job.

Martini991:
I wouldn't assume the first point, there are many ways someone can approach
the module! For example, it's common to do a research before starting a new
project and choose the technology based on modules/libraries available for
it. If this functionality is what someone needs, they can test it STAT and
choose Django over something else, all this much before any code is laid
down.

And again, more people, more contributions! 😉

Also, installing docker over two DB have a huge difference, not only an
apt-get command! Examples: what if I have already installed one or more
DBs? I have to extract/find/remember the credentials and do all the work to
configure the connections in the settings? And if I don't want to use those
instances because there is running something important and I fear this
module can brake it or simply don't want to "put dirty" in them? Does my OS
have all the necessary libs to build the required python modules (posgres
and mysql drivers are non-trivial)?

There is many points where someone can fall, but download docker, git clone
the repo, and launch ./docker-run.sh from within the project root dir is
straight forward, batteries included and clean.

And I still don't know the answer of my two previous questions: why you
don't want to use it? Are really 5 text files of about 130 lines in total a
burden to understand?

I don't know, I feel like I'm trying to sell you a vacuum cleaner you don't
want, but in reality I have spent time to solve a problem someone else can
have, and I would be happy to not make them solve it twice... also I'm not
selling anything, it's free! 😆

I'll renew the proposal of a Skype, because I feel it will be orders of
magnitude more efficient compared to GitHub-comment-texting*™.*

In any case, have a nice day! 😄

PR #838, opened 2020-06-03
Added ability to change transaction rollback conditions.

Tartorus:
Sometimes it is necessary to save changes in models and return a response with a status code of 400. In my case, we check the SMS code from the client and send the number of remaining attempts if it is not correct.


etianen:
Thanks for the MR!

I'm not convinced that this use-case isn't better served by your own RevisionMiddleware in your own codebase. It's pretty easy to implement using reversion.create_revision(), and avoids complicating the main django-reversion codebase.

PR #842, opened 2020-08-15
Add initial implementation for use_natural_foreign_keys support

### LINKED ISSUE #841: Use natural key for serialization
Hi,

Is it any possible to use [natural keys](https://docs.djangoproject.com/en/3.0/topics/serialization/#natural-keys) when serializing relations?

I see that default Django `serializers.serialize()` is used for serializing, but there is no way to pass `use_natural_foreign_keys=True` or `use_natural_primary_keys=True`.

matwey:
Close #841


etianen:
Really nice work, thanks!

etianen:
I'll release this in about a week, if no other features land before then.

Issue #840, opened 2020-08-07
Deprecation warning for django 4.0


GitRon:
Hi there,

I just saw a deprecation warning:

> site-packages\reversion\admin.py:142: RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path().
 
I think this might affect the smalles django version this project supports. Still, seems to be worth having a look at.

Best regards
Ronny

etianen:
Happy to take a MR to address this. Probably dropping older versions of
Django is best than shimming django-reversion, so long as any Django
versions we drop are no longer supported by Django itself.

GitRon:
@etianen I'm quite busy at the moment but I'll see what I can do

GitRon:
@etianen I saw you already fixed the url thing. I found the signal args deprecation and fixed it and opened a PR. 

`ugettext` was also fixed in the meantime so we might be good to close this issue afterwards.

etianen:
Thanks for your work on this. :)

It's not really worth a new release for fixing a deprecation warning, but it'll be folded into the next release.

PR #846, opened 2020-09-01
Removed `providing_args` argument from signals because of django 4.0 …

GitRon:
…deprecation. Also removed `_signal_args` var from signals.py

Fixes https://github.com/etianen/django-reversion/issues/840


etianen:
Thanks!

PR #847, opened 2020-09-02
Run tests against Django 3.1

michael-k:
- `databases` was introduced in Django 2.2, see https://docs.djangoproject.com/en/dev/releases/2.2/#tests
- `multi_db` was deprecated at the same time, see https://docs.djangoproject.com/en/dev/releases/2.2/#id2


etianen:
Thank you!

PR #852, opened 2020-10-16
Add existing meta data instance

jsoa:
## What

The code in this change set introduces the ability to add an existing meta data instance to ``add_meta`` so that we can create meta data instances before a reivision is created.

## Why

In certain scenarios, when creating revisions manually, we may want to create a meta data instance before a revision is created. 

For example

*models.py*
```python
class MyModel(models.Model):
    name = models.CharField(max_length=200)


class MyRevisionMeta(models.Model):
    revision = models.OneToOneField(Revision, null=True, on_delete=models.CASCADE)
    version_label = models.CharField(max_length=20)
    mymodel = models.ForeignKey(MyModel, on_delete=models.CASCADE)
```
*views.py*

```python
def create_mymodel(request):

    inst = MyModel.objects.create(name='')
    MyRevisionMeta.obects.create(mymodel=inst, version_label='original')

    return ...
```

```python
def update_mymodel(request):

    # Normal updates (no revisions)

    return ...
```

```python
def publish_mymodel(request):

    with reversion.create_revision():
        inst = MyModel.objects.get(pk=...)
        inst.name = 'name change'
        inst.save()
        meta = MyRevisionMeta.objects.get(mymodel=inst)
        reversion.add_meta(meta)
   ....
```

## How

Updates to ``add_meta`` was changed to allow for either `model` or `model instance`. Specifically, the ``create`` that was done before was changed to a ``update_or_create``. i.e.

```python
        meta_class._base_manager.db_manager(
            using=using
        ).update_or_create(
            pk=pk,
            defaults=meta_fields
        )
```

With this change we supply ``update_or_create`` a ``pk``, which will either be the ``pk`` from the ``instance`` or ``None`` when a normal ``model`` is supplied.

And in order for the ``revision`` instance to be updated as well as created, it was added to the ``meta_fields`` dict and supplied as ``defaults`` to ``update_or_create``


etianen:
Hi!

I'm not sure this approach is a good idea. What advantage could there be in saving a metadata model twice (once to create, again to update_or_create)? It turns one database query into two. Worse, a user might re-use an existing metadata instance, expecting it to be associated with two revisions, when in reality the metadata instance will be re-associated with the new revision, and removed from the old.

jsoa:
@etianen thanks for the feed back, our current implementation/work-around for this is to have a separate model that has a nullable FK to our meta data model, and this was an attempt to eliminate the extra model. 

Here are some possible solutions we could try to address your concerns:

- With some tweaks to the code we could make the logic only do the one query, one to create when a ``Model`` is supplied or an update when an ``instance`` is supplied.  The ``update_or_create`` was me trying to keep the code slim.

- Further more we could also check to make sure, when an ``instance`` is supplied, that the required ``reivison`` FK is null before attempting the update, which would address the re-associating problem.

etianen:
Can I ask why you're creating your meta model before the revision is
created? Surely you could just store all the information in a dict, then
use reversion.add_meta(YourModel, **your_model_data)?

By creating it before the revision is saved, you're performing two database
queries: one to create, then one to update.

jsoa:
Surely, and you are correct, we do store the data for later use, in the form of an extra model however, with a `OneToOneField` to a custom `RevisionMeta` model. 

### More information:
We wanted a way to create revisions manually, based off of a certain status change of a database record, and reversion gave us that functionality. So there might be many normal saves between a certain status change that we did not want to store revisions. During this process, the edits might take a couple hours to perhaps months and we wanted a way for our users to "tell us about the change your doing" before or during normal updates and before submitted/publishing the data (which triggers a status change and therefore a `revision` to be made).

Here is our current work around

```python
class RevisionExtra(models.Model):
    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
    object_id = models.CharField(max_length=255)
    content_object = GenericForeignKey('content_type', 'object_id')

    version_number = models.CharField(max_length=255) # user input
    version_date = models.DateField(default=now) # user input
    # extra user input fields


class RevisionMeta(models.Model):
    revision = models.ForeignKey(Revision, on_delete=models.CASCADE)
    revision_extra = models.OneToOneField(RevisionExtra, on_delete=models.CASCADE)
```

We store all the real meta data in ``RevisionExtra``, and during the user "editing" a content object, they will have the ability to edit `RevisionExtra`, in this case a `version_number` and `version_data`. These edits of a content object might take a while before they "publish" the changes, which is when we manually create the `Revision` and attach our `RevisionExtra` to it.

Here is some of our workflow (using django rest framework)

```python
class ObjectViewSet(viewsets.ViewSet):

    # extra view stuff

    def update(self, request, ...):
        # normal updates

    def create(self, request):
        # create objects

    @action(detailt=True, methods=['PUT']
    def start_edit(self, request, ...):
        """specific route for initiating an edit"""
        obj = self.get_object()
        obj.status = 'in_progress'
        obj.save()

        RevisionExtra.objects.create(
            content_type=ContentType.objects.get_for_model(obj),
            object_id=obj.id,
            version_number=f'edit {str(uuid4())[:4]}' # auto generate number initially
         )
        return Response(...)
        
    @action(detail=True, methods=['PUT']
    def publish(self, request, ...):
        """specific route for publishing"""
        obj = self.get_object()

        with reversion.create_revision():
            obj.status = 'published'
            obj.save()

            c_type = ContentType.objects.get_for_model(obj)
            revision_extra = RevisionExtra.objects.get(
                revisionmeta__isnull=True,
                content_type=c_type,
                object_id=str(obj.id)
             )

            # Store some meta-information.
            reversion.set_user(request.user)
            reversion.add_meta(RevisionMeta, revision_extra=revision_extra)
        return Response(...)
```

The idea for this PR was to eliminate the need for the extra model.

*For what its worth, we took this approach to only store revisions of data that matters to us (published versions of data), as well as keeping our database storage somewhat leaner.*

etianen:
I see. That does make sense. Thank you for clarifying. :)

I think your extra model is a perfectly decent approach. It's a little less
efficient to look up, but it's also a little more efficient to save the
revision. Either way, I doubt the difference is noticeable, and I'd rather
keep django-reversion leaner and less complex for the common-case.

jsoa:
thx for the feedback, much appreciated

PR #866, opened 2021-04-07
Run tests against Django 3.2

michael-k:
And add missing trove classifier for Python 3.9.


etianen:
Thanks!

Interestingly, no actions seemed to run for your MR. Any idea why? Is there something I need to turn on?

michael-k:
> Interestingly, no actions seemed to run for your MR. Any idea why? Is there something I need to turn on?

`on: [push, pull_request]` instead of `on: [push]`

https://github.com/etianen/django-reversion/blob/7f3645649bcd92d832cb69754fded6e9a8802aa6/.github/workflows/python-package.yml#L3

Issue #873, opened 2021-06-04
Django 3.2.4 Upgrade - DEFAULT_AUTO_FIELD


leonardodepaula:
After upgrading to Django 3.2.4 I started to receive those warning messages:

WARNINGS:
reversion.Revision: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
reversion.Version: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.

How can I set default_auto_field for django-reversion?

Thanks.

etianen:
This warning is harmless for Django 3.2. This will be fixed in the next
reversion release. Alternatively, I'll take a MR to fix it now.

leonardodepaula:
Thanks, @etianen.

Peter-van-Tol:
Done, hopefully it can be part of the next release!

etianen:
Definitely :)

PR #876, opened 2021-06-16
Create apps.py

Peter-van-Tol:
Solves #873.

By adding an AppConfig, one can set the `default_auto_field`. The other parameters of the AppConfig are chosen so the behaviour for other installations remains unchanged.

### DIFF
--- a/reversion/apps.py
+++ b/reversion/apps.py
@@ -1,5 +1,6 @@
 from django.apps import AppConfig
 
+
 class ReversionConfig(AppConfig):
     name = 'reversion'
-    default_auto_field = 'django.db.models.BigAutoField'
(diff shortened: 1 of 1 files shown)


etianen:
Interesting. I think BigAutoField is the correct choice here, but won't that also require a migration to be added?

I think for the time being keeping it as AutoField is safer. I've found some bugs in the migration Django generates for this in some of my work projects.

Peter-van-Tol:
In the PR I've changed the BigAutoField to AutoField. Also, the workflow was cancelled due to Flake8 error. This has been resolved as well.

etianen:
Merged, thanks!

The tests are failing for an unrelated issue: https://github.com/psycopg/psycopg2/issues/1293#issuecomment-862835147

PR #897, opened 2021-10-29
Python 3.10: run automated tests for Django 4.0.x against Python 3.10

kevinmarsh:
[Django >4.0.x supports Python 3.10](https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django) so would be good to test against it. All the tests passed locally (Django==4.0.0b1, Python 3.10.0) with postgres but didn't test mysql.

Thought it was better to make it not very DRY and explicitly _exclude_ the older Django versions then having a single _include_ for 3.10 and forgetting to add another include when 4.1.x is released.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -41,7 +41,7 @@ jobs:
       DJANGO_DATABASE_PASSWORD_MYSQL: root
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


etianen:
Brilliant, thanks!

PR #898, opened 2021-11-01
Improved performance of `createinitialrevisions` management command on postgres databases

philipstarkey:
Rather than retrieving the list of PKs that needs revisions created for, and then pulling out the actual objects in separate queries using `in_bulk`, we drop it down to a single `count()` query (to get the total number) and an `iterator()` query. This improves the performance (and possibly memory usage) on postgres databases when server side cursors are enabled (which they are by default on postgres databases). For databases where server side cursors are not enabled, we revert to the original code as `iterator()` is memory intensive (at the database driver level, see Django documentation on `iterator()` for further details).

My local tests suggest that this provides a speed-up of a factor of 2. Would be good if someone else could verify to confirm that's not just an artefact of my set-up though!

### DIFF
--- a/reversion/management/commands/createinitialrevisions.py
+++ b/reversion/management/commands/createinitialrevisions.py
@@ -2,7 +2,7 @@ import json
 
 from django.apps import apps
 from django.core.management import CommandError
-from django.db import reset_queries, transaction, router
+from django.db import connections, reset_queries, transaction, router
 from reversion.models import Revision, Version, _safe_subquery
 from reversion.management.commands import BaseRevisionCommand
 from reversion.revisions import create_revision, set_comment, add_to_revision, add_meta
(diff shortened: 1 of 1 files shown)


etianen:
This is very nice, and the extra complexity it introduces is worth it for the performance gain.

Can you extend the tests for this command to test on mysql, postgres and sqlite by passing the "using" param to the management command? That will ensure that both codepaths are tested.

philipstarkey:
@etianen Apologies for the delay, I completely spaced on the existence of this PR.

There were already tests that used postgres and mysql [here](https://github.com/etianen/django-reversion/blob/dbf87494ef651ca31802e2566bac708871c6e53d/tests/test_app/tests/test_commands.py#L54-L67). Would it be sufficient to just add an additional test for sqlite, or would you like something further?

etianen:
Oh, good spot!

Yes, an SQLite test would be nice. Although by the looks of things the MySQL test is already covering both code paths. So it's not as essential as I first thought.

If you fancy adding an SQLite test, please do! Otherwise, I think this is good to merge as-is.

etianen:
Huh, sorry, this slipped below my radar. Merged!

PR #904, opened 2022-02-04
Support asyncio views

bellini666:
Use contextvars instead of threading.local to support async views.

I tried to do the smallest changes that I could, but using contextvars that code can be simplified a lot.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -41,7 +41,7 @@ jobs:
       DJANGO_DATABASE_PASSWORD_MYSQL: root
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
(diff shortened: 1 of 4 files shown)


etianen:
Hi,

Thanks for this! It's a good change. Unfortunately, it requires Python 3.7+, which is going to be supported by Django LTS until early 2024!

I'd like to keep this MR open until then, but I don't think I can drop support for Python 3.6 while Django LTS supports its.

Alternatively, I'd consider a compatibility shim that used contextvars and fell back to thread-local state. That seems like it would be a relatively unobtrusive change.

etianen:
You make an excellent point. Go ahead and drop 3.6 from the matrix and
we'll get this merged. :)

bellini666:
@etianen nice :)
Just made the change, think it should be enough

etianen:
Thanks! I'll get this into a release ASAP.

PR #912, opened 2022-06-16
Fix admin detail view in multi-database environment

atten:
There's a bug with false-positive trigger of **Cannot use VersionAdmin with a database that does not support savepoints.** exception on the admin page of a versioned model. It can be reproduced with Django configuration with multiple databases:

```
# default DB backend is dummy
DATABASES = {
    'default': {},
    'db1': {
        'ENGINE': 'django.db.backends.postgresql',
        ...
    },
    'db2': {
        'ENGINE': 'django.db.backends.postgresql',
        ...
    }
}
```

Thus, `connection.features.uses_savepoints` is irrelevant and should be replaced with `connections[version.db].features.uses_savepoints`.

### DIFF
--- a/reversion/admin.py
+++ b/reversion/admin.py
@@ -1,5 +1,5 @@
 from contextlib import contextmanager
-from django.db import models, transaction, connection
+from django.db import models, transaction, connections
 from django.contrib import admin, messages
 from django.contrib.admin import options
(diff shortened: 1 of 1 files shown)


etianen:
Good catch!
The failed builds were due to an unrelated problem with Sphinx

etianen:
Released as 5.0.1

PR #917, opened 2022-08-10
Add support of Django 4.1

leo-naeka:
Hi ! This PR introduces support of Django 4.1

Also removes support of deprecated Django versions (all of those below 3.2 LTS).

I've run both pyupgrade and django-upgrade and isolated these changes in separate commits. Feel free to squash them.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -43,28 +43,16 @@ jobs:
       matrix:
         python-version: [3.7, 3.8, 3.9, '3.10']
         django-version:
-          - '>=4.0a1,<4.1'
+          - '>=4.1,<4.2'
+          - '>=4.0,<4.1'
           - '>=3.2,<4.0'
-          - '>=3.1,<3.2'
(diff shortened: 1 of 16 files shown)


etianen:
This is fantastic, a really thorough job.

I'll not bother with a release from this, as nothing has actually _changed_. But this'll get folded into the next release.

leo-naeka:
> This is fantastic, a really thorough job.

Great! Most of the thorough job was performed by upgrading tools actually 😉

I would be very pleased to open another PR to add a [pre-commit](https://pre-commit.com/) config with a few hooks like pyupgrade and django-upgrade. From that, you'd be able to setup the [Github integration](https://pre-commit.ci/) and eventually add other hooks like black, isort, flake8…

Let me know!

etianen:
Thanks for the offer, but I'm not really a fan of pre-commit. I don't like lag introduced into my commits, and prefer to use IDE integrations of these tools along with CI checks to enforce before merging.

PR #926, opened 2022-10-26
Update Django version required in README to 3.2

andyzickler:
The minimum version of Django was updated in https://github.com/etianen/django-reversion/pull/917 but the readme still showed 2.0 as supported.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,7 +12,7 @@ Requirements
 ============
 
 - Python 3.7 or later
-- Django 2.0 or later
+- Django 3.2 or later
 
 Features
 ========


andyzickler:
Hmm, It looks like this change happened in release 5.0.3? Should a note be added to the changelog too? I can add one to this pr if you'd like.

etianen:
Good spot!

I'm not really fussed about the changelog. Dropping support for out-of-support Django versions is a non-event, IMO.

daulam:
Hi @etianen, The PyPI page is still out-of-date in terms of the Django requirement for 5.0.3 - https://pypi.org/project/django-reversion/5.0.3/ 

(And just my opinion) About the changelog, it acts as log of notable events/changes made to a project. Sure Django 2.0 is very old but still for historical completeness, timeline and previous patterns in the changelog - I feel it's worthwhile to note it because it's a major change in requirements. Also maybe worthwhile to note the support for Django 4.1. For me, I was wondering when this requirement was bumped because of the discrepancy on the PyPI page; tried looking at the changelog file and no info until I landed here.

etianen:
gI'm not opposed to putting this in the changelog. I'm just not
particularly excited about it either. If you write it, I'll merge it. :P

The PyPI page, AFAIK, will only update with a new release. Or is there a
way to update the description and metadata for an already-released package?

PR #928, opened 2022-11-21
Introduce default format setting.

proofit404:
Good evening,

The reason I'm interested in this pull request to be landed:

We frequently use model field from third-party library which can't be handled by core Django serializers.

I wrote custom serializer, not big deal. But now I need to mention custom format name in every call to `register` decorator.

Would be nice to set it once and forget about it :smile: 

Best regards,
Josiah.

### DIFF
--- a/reversion/revisions.py
+++ b/reversion/revisions.py
@@ -5,6 +5,7 @@ from functools import wraps
 from django.apps import apps
 from django.core import serializers
 from django.core.exceptions import ObjectDoesNotExist
+from django.conf import settings
 from django.db import models, transaction, router
 from django.db.models.query import QuerySet
 from django.db.models.signals import post_save, m2m_changed
@@ -366,7 +367,7 @@ def _get_senders_and_signals(model):
         yield m2m_model, m2m_changed, _m2m_changed_receiver
 
 
-def register(model=None, fields=None, exclude=(), follow=(), format="json",
+def register(model=None, fields=None, exclude=(), follow=(), format=None,
              for_concrete_model=True, ignore_duplicates=False, use_natural_foreign_keys=False):
     def register(model):
         # Prevent multiple registration.
@@ -388,7 +389,7 @@ def register(model=None, fields=None, exclude=(), follow=(), format="json",
                 if field_name not in exclude
             ),
             follow=tuple(follow),
-            format=format,
+            format=format or getattr(settings, "REVERSION_DEFAULT_FORMAT", "json"),
             for_concrete_model=for_concrete_model,
             ignore_duplicates=ignore_duplicates,
             use_natural_foreign_keys=use_natural_foreign_keys,
--- a/reversion/revisions.py
+++ b/reversion/revisions.py
@@ -2,6 +2,7 @@ from contextvars import ContextVar
 from collections import namedtuple, defaultdict
 from contextlib import contextmanager
 from functools import wraps
+from types import SimpleNamespace
 from django.apps import apps
 from django.core import serializers
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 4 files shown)


etianen:
Hi! This is an excellent PR, and I appreciate you taking the time to add documentation and tests. :+1: 

I have one concern with this - historically I've avoided adding any settings to `django-reversion`, since there is only one entry-point to configuration (`reversion.register`). What you're trying to do here (default settings for `reversion.register`), could be acheived by wrapping `reversion.register` with your own register function. If you use the admin integration, you can likewise override the `reversion_register` in a common base class.

That being said, I don't want to be dogmatic about this. If you think there's an argument for adding the first `django-reversion` setting, and it suits your own use-case better, then I'm happy to change my mind. Or maybe you hadn't considered wrapping the `register` function, and prefer that approahc. I'm very happy to discuss it further here, and happy to be pursuaded! :P

If we do go ahead with adding a setting, this PR is very ready to merge, and I've no requested changes. :heart:

proofit404:
Hello,

We have about 300 models decorated with vanilla `@register` call.

Intention behind this pull request was my confusion.

There is no easy way to change default behavior in the existing code base.

We were forced to touch 300 decorator calls.

Be it either new value for the argument or new decorator function.

None of these changes does not follow DRY principle.

Of course, it's only a nice to have option if someone would encounter similar situation.

If you agree to introduce settings to the library, I would like to spend a little bit more time on this PR.

First of all, it could be a good idea to make settings variable a mapping similar to approach in DRF.

Also, I would like to write a little tutorial how we solved our problem from cover to cover with datamigration for existed model instances.

What do you think?

Best regards,
Josiah.

proofit404:
Any decision?

etianen:
Sorry, yes. The settings dict is a good idea, as is the idea of a default
format setting.

etianen:
I would merge that

proofit404:
Good afternoon,

I added application settings class to handle user settings module for all register arguments.

If you are happy with intended design and choisen implementation, I'll start writting documentation for it.

Thanks!

reversion/revisions.py
```diff
@@ -407,6 +414,19 @@ def register(model):
...
+def _parse_register_arguments(fields, exclude, follow, format, for_concrete_model, ignore_duplicates,
+                              use_natural_foreign_keys):
+    return SimpleNamespace(
+        fields=fields or app_settings.DEFAULT_FIELDS,
+        exclude=exclude or app_settings.DEFAULT_EXCLUDE,
+        follow=tuple(follow or app_settings.DEFAULT_FOLLOW),
+        format=format or app_settings.DEFAULT_FORMAT,
+        for_concrete_model=for_concrete_model or app_settings.DEFAULT_FOR_CONCRETE_MODEL,
+        ignore_duplicates=ignore_duplicates or app_settings.DEFAULT_IGNORE_DUPLICATES,
+        use_natural_foreign_keys=use_natural_foreign_keys or app_settings.DEFAULT_USE_NATURAL_FOREIGN_KEYS,
```
etianen, line 426: This shortcut logic isn't quite right. You want to use the pattern:

```
fields=app_settings.DEFAULT_FIELDS if fields is None else fields 
```

Otherwise, an empty string or tuple will count as "unset"

etianen (changes requested):
A very nice design and implementation.

I've made one comment in the diff. The other change I'd request is to drop the following:

- `DEFAULT_FIELDS`
- `DEFAULT_EXCLUDE`
- `DEFAULT_FOLLOW`

Those register arguments are not something you'd ever want to set globally.

reversion/revisions.py
proofit404: Fixed :+1:

proofit404:
If there is no outstanding code issues, I would starting to write documentation.

etianen:
I'm very happy with the code! Please do go ahead with docs.

umairahmed515:
@etianen when are we going to merge this PR?

proofit404:
I didn't updated pull request yet.

We have a newborn, so I highly limited in time.

I'm still interested to finish this contribution.

PR #932, opened 2022-12-29
Set default auto field in tests.

proofit404:
# Before
Untitled
# After
Untitled1


etianen:
Brilliant, thank you!

PR #941, opened 2023-05-15
Add Django 4.2 and Python 3.11 to the CI

browniebroke:


etianen:
Lovely! I'll not bother with a release for this, but it's nice to keep the CI up-to-date!

PR #947, opened 2023-09-28
Fix “has no attribute” error

julianklotz:
`object_repr` is an attribute of the version, not the admin instance.


etianen:
Whoops

PR #950, opened 2023-10-12
Move dropping Python 2.7 changelog to correct version

chicheng:
Sorry the previous pull request added the log to the wrong version, also adds the author name according to https://github.com/etianen/django-reversion/pull/801


etianen:
👍

PR #951, opened 2023-11-07
fix: speed up get deleted. remove unnecessary subquery

caullla:
This subquery is not nessesary and only cause huge load and slow recovery page load time.


etianen:
Very nice! Release incoming.

PR #955, opened 2023-12-19
perf: Fixes N+1 queries while rendering the recover_list.html template

armonge:
While running a project with django-reversion and Sentry we found out there's one template that makes a query to the DB on every iteration. This PR fixes that

https://docs.sentry.io/product/issues/issue-details/performance-issues/n-one-queries/
https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.select_related


etianen:
Very nice, I'll release shortly.

PR #960, opened 2024-01-30
Add missing migration

pinguin999:


etianen:
Thanks! I'm going to merge this to give you credit for a fast fix, but then go with the solution in #961.

PR #969, opened 2024-06-06
Switch call from removed is_hidden method to hidden property

jeremy-engel:
`ForeignObjectRel.is_hidden()` was removed in
https://github.com/django/django/pull/17862 in favor of the `hidden` property of the same class. This change will be released in django 5.1. In previous versions of django, the `hidden` property called the `is_hidden()` method, so there should not be changes in behavior for older versions.


etianen:
Thanks!

PR #971, opened 2024-06-17
Remove irrelevant comment

GeyseR:


etianen:
I was about to say something snooty about it being a harmless and fun comment. But you're right, it's actually an incorrect comment. So away it goes! ❤️

PR #972, opened 2024-07-19
Re-introduce bulk_create on supported databases

### LINKED ISSUE #924: Re-add support for `bulk_create`
Between [1.8.0](https://django-reversion.readthedocs.io/en/stable/changelog.html#id50) and 1.8.6 (#380), reversion used `bulk_create` when creating versions, which presumably dramatically improved performance when creating a revision with many changed model instances[1]. It sounds like the motivation for removing this was inability to get back object primary keys from `bulk_create`, but that has since been fixed on some popular databases -- [Django 1.10](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#bulk-create) (released about a year and a half after #380) added some support (only PostgreSQL at the time). SQLite was added in [Django 4.0](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#bulk-create), and MariaDB was added sometime in the middle. It looks like there's a convenient feature test for this (`connection.features.can_return_rows_from_bulk_insert`), so it would be nice if django-reversion could use `bulk_create` when feasible.

(I found #380  while trying to figure out why tons of queries were being used despite the [changelog](https://django-reversion.readthedocs.io/en/stable/changelog.html#id50) claiming reversion uses bulk_create, so if you don't re-add `bulk_create` support it might be nice to note the change somewhere?)

[1] I haven't benchmarked this specifically, but a script that creates several thousand objects went from 30 minute runtime to ~3 minutes by doing far more query batching, so I suspect this is significant.

stianjensen:
This fixes #924.

I see the test suite already runs with both postgres and mysql, so in that regard both of these branches should already get covered across the existing test suite. Let me know if it makes sense to add more tests around this switching specifically!

### DIFF
--- a/reversion/revisions.py
+++ b/reversion/revisions.py
@@ -5,7 +5,7 @@ from functools import wraps
 from django.apps import apps
 from django.core import serializers
 from django.core.exceptions import ObjectDoesNotExist
-from django.db import models, transaction, router
+from django.db import models, transaction, router, connections
 from django.db.models.query import QuerySet
 from django.db.models.signals import post_save, m2m_changed
 from django.utils.encoding import force_str
@@ -211,6 +211,7 @@ def add_to_revision(obj, model_db=None):
 
 def _save_revision(versions, user=None, comment="", meta=(), date_created=None, using=None):
     from reversion.models import Revision
+    from reversion.models import Version
     # Only save versions that exist in the database.
     # Use _base_manager so we don't have problems when _default_manager is overriden
     model_db_pks = defaultdict(lambda: defaultdict(set))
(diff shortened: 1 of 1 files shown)


reversion/revisions.py
```diff
@@ -248,9 +249,17 @@ def _save_revision(versions, user=None, comment="", meta=(), date_created=None,
...
     revision.save(using=using)
     # Save version models.
+
+    can_use_bulk_create = connections[using].features.can_return_rows_from_bulk_insert
+
     for version in versions:
         version.revision = revision
-        version.save(using=using)
+        if not can_use_bulk_create:
+            version.save()
```
etianen, line 258:
```suggestion
            version.save(using=using)
```

etianen (commented):
This looks excellent! Can you review my suggested change to see if it makes sense?

reversion/revisions.py
stianjensen: Ah, ofc! Good catch 😅

etianen:
Merged, thank you@

etianen:
Will shortly be released as v5.1.0

PR #976, opened 2024-10-30
Add ability to use RevisionMixin on async views.

tbeadle:


bartvanandel:
Will this also resolve any `Asynchronous handler adapted for middleware reversion.middleware.RevisionMiddleware.` warnings? Comparing [`RevisionMiddleware`](../tree/master/reversion/middleware.py) to e.g. [`CorsMiddleware`](https://github.com/adamchainz/django-cors-headers/blob/main/src/corsheaders/middleware.py), it seems to me that it may also need some adaptations?

Issue #963, opened 2024-02-24
Use object_id other than primary key


sourabhv:
I am working on cleaning an old database where the revisions of data are kept in same table with different `id` PK but share the same `uuid` (not the PK). When moving old data into django reversion before deleting it from this table, it picks up `id` and does not map values based on `uuid`. Is there a way to tell reversion to use a column besides `pk`?

etianen:
This isn't supported right now, but could be with an additional parameter passed to `reversion.register()`. I'd happily take a PR for it.

PR #999, opened 2026-05-02
Feature: add custom object id field 

philippeducasse:
from issue #963

Adds an object_id_field parameter to register() that allows versioning objects using a custom field (e.g. a UUID or slug) as the stored object_id instead of the primary key. This is useful when migrating legacy data where the natural identifier is of the database pk. 

limitation: get_deleted() still assumes object_id == pk and does not support object_id_field. This can be addressed in a follow-up.


reversion/revisions.py
```diff
@@ -376,7 +388,7 @@ def _get_senders_and_signals(model):
 
 
 def register(model=None, fields=None, exclude=(), follow=(), format="json",
-             for_concrete_model=True, ignore_duplicates=False, use_natural_foreign_keys=False):
+             for_concrete_model=True, ignore_duplicates=False, use_natural_foreign_keys=False, object_id_field=None):
```
etianen, line 384: It would be more straightforward to make `object_id_field` default to `pk` here. Then you don't need to handle the `None` case anywhere else in the code.

reversion/revisions.py
```diff
@@ -168,7 +169,10 @@ def _add_to_revision(obj, using, model_db, explicit):
         return
     version_options = _get_options(obj.__class__)
     content_type = _get_content_type(obj.__class__, using)
-    object_id = force_str(obj.pk)
+    if version_options.object_id_field:
```
etianen: By making `register` default to `pk` rather than `None`, we can completely remove this `if` statement.

reversion/revisions.py
```diff
@@ -209,6 +215,10 @@ def add_to_revision(obj, model_db=None):
         _add_to_revision(obj, db, model_db, True)
 
 
+def _get_object_id_field(model):
```
etianen: Why are you checking for `is_registered` here? Is it ever correct to call `_get_object_id_field` for an unregistered model? If not, the whole `if is_registered(model) else "pk"` branch is redundant.

reversion/models.py
```diff
@@ -135,6 +138,12 @@ def get_for_object_reference(self, model, object_id, model_db=None):
         )
 
     def get_for_object(self, obj, model_db=None):
+        if is_registered(obj.__class__):
```
etianen: By making `object_id_field` default to `pk`, this whole `if` statement can be removed.

reversion/admin.py
```diff
@@ -48,6 +48,22 @@ def reversion_register(self, model, **kwargs):
         """Registers the model with reversion."""
         register(model, **kwargs)
 
+    def get_reversion_object_id(self, request, object_id):
```
etianen: I don't really see why these methods exist? Can they not be automatically handled by looking up `object_id_field` and retrieving the appropriate model field?

etianen (changes requested):
Not a bad start, and the tests are appreciated!

I've added some suggestions to simplify it.

I'd also like to have `get_deleted` and any other functionality updated to respect this new registration option before merging anything,.

philippeducasse:
Hi @etianen ! thanks for the feedback and suggestions. Ive refactored the code, hopefully its more logical now. I also refactored the get_deleted functionality. I hope I haven't forgotten anything!

reversion/revisions.py
```diff
@@ -209,6 +212,11 @@ def add_to_revision(obj, model_db=None):
...
```
etianen: There are some parts of the codebase using `VersionOptions.object_id_field` directly, and others using `_get_object_id_field()`.

What if... we resolved `object_id_field` to a concrete field name in `register`, stored _that_ as `VersionOptions.object_id_field`, and then used `VersionOptions.object_id_field` consistly everywhere, dropping this helper?

So make `object_id_field` default to `None`. If it's `None`, we resolve it to a concrete primary key field using `._meta.pk.name`. If it's a string, we validate it using `.get_field` during register.

This means explicitly setting `object_id_field="pk"` becomes an error.

reversion/admin.py
```diff
@@ -200,7 +200,13 @@ def _reversion_revisionform_view(self, request, version, template_name, extra_co
...
                 # Run the normal changeform view.
                 with self.create_revision(request):
-                    response = self.changeform_view(request, quote(version.object_id), request.path, extra_context)
+                    obj = get_object_or_404(
+                        version._model._default_manager.using(version.db),
+                        **{_get_options(version._model).object_id_field: version.object_id},
+                    )
+                    response = self.changeform_view(
+                        request, quote(str(obj.pk)), request.path, extra_context
+                    )
```
etianen, line 214: This feels worth optimizing for the case of `object_id_field` being the primary key. It'll save one query for 99.9% of django-reversion users.

So a check like `._meta.pk.name == object_id_field`

reversion/admin.py
```diff
@@ -305,6 +311,8 @@ def history_view(self, request, object_id, extra_context=None):
             if not self.has_change_permission(request):
                 raise PermissionDenied
 
+        obj = get_object_or_404(self.model, pk=unquote(object_id))
+        reversion_object_id = str(getattr(obj, _get_options(self.model).object_id_field))
```
etianen: As above, this is worth optimizing for the common case of `object_id_field` being the primary key.

etianen (changes requested):
Thanks for your changes. This feels a lot cleaner.

I've suggested one simplification, and a couple of admin optimizations. We're pretty close.

I'd also ask:

- You add this new `register` field to the docs.
- Grep the codebase for explicit uses of `pk` to make sure you've not missed anything.

reversion/revisions.py
philippeducasse: resolving object_id_field here using ._meta.pk.name resulted in errors, for models with multi-table inheritence, so instead i used meta.pk.attname. 

```
if object_id_field is None:
            id_field = model._meta.pk.attname
        else:
            model._meta.get_field(object_id_field)
            id_field = object_id_field
```

reversion/admin.py
philippeducasse: great yes thanks for catching this

philippeducasse:
Added the parameter to docs, and also greped the codebase :

grep -n "obj\.pk\|get_object_or_404.*self\.model\|getattr.*object_id_field"

the only matches are in admin.py and revisions.py: Lines 211 and 323–324 in admin.py are both in else branches; the slow path that only runs when  object_id_field != pk.attname. The two hits in revisions.py are unrelated: line 168 is an unsaved-object  guard, line 394 is the registration resolution itself.

etianen (approved):

etianen:
This is great! Thanks for your patience - I'll get this released tonight.

etianen:
Released v6.2.0

